PR #15, opened 2017-01-17
Create new job to trigger nightly build using Mock arguments

lisibann:
Nightly builds run build_host_os using Mock arguments to enable tests (check phase).

### DIFF
--- a/jenkins_jobs/trigger_nightly_host_os_build/post_build_script.sh
+++ b/jenkins_jobs/trigger_nightly_host_os_build/post_build_script.sh
@@ -1,0 +1,15 @@
+NIGHTLY_DIR_NAME=$(cat NIGHTLY_DIR_NAME)
+BUILD_DIR_PATH="../to_build/${TRIGGERED_BUILD_NUMBER_build_host_os}"
+
+# update nightly build dir and latest nightly build
+ln -s "$BUILD_DIR_PATH" "$NIGHTLY_DIR_NAME"
+ln -s "$NIGHTLY_DIR_NAME" latest
+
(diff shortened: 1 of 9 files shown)


jenkins_jobs/trigger_nightly_host_os_build/pre_build_script.sh
```diff
@@ -0,0 +1,3 @@
+RELEASE_DATE=$(date +%Y-%m-%d)
+echo "$RELEASE_DATE" > NIGHTLY_DIR_NAME
+echo "$TRIGGERED_BUILD_NUMBER_build-host-os"
```
ghost: Does this var has any value? The build has not been triggered yet at this point. Also, the job names have underscores instead of hyphens.

jenkins_jobs/trigger_nightly_host_os_build/post_build_script.sh
```diff
@@ -0,0 +1,15 @@
+NIGHTLY_DIR_NAME=$(cat NIGHTLY_DIR_NAME)
+BUILD_DIR_PATH="../to_build/${TRIGGERED_BUILD_NUMBER_build_host_os}"
```
ghost, line 2: It's better to add the UPLOAD_SERVER_BUILDS_DIR parameter to use here.

jenkins_jobs/trigger_nightly_host_os_build/pre_build_script.sh
lisibann: Right. It does nothing at this point. We have the same for trigger_weekly. I'll remove from both.

ghost (approved):

PR #16, opened 2017-01-24
Change build upload directory name from build number to timestamp

ghost:
When setting up a new Jenkins instance, we had to be careful to
configure the initial build in a way that it would not overlap with
previously uploaded builds.

By using a timestamp we make sure the overlaps will be more
unlikely to happen.

### DIFF
--- a/jenkins_jobs/build_host_os/script.sh
+++ b/jenkins_jobs/build_host_os/script.sh
@@ -1,6 +1,9 @@
+# ISO 8601 date with nanoseconds precision
+TIMESTAMP=$(date --utc -Ins)
 VERSIONS_REPO_DIR=$(basename $VERSIONS_REPO_URL .git)
 MOCK_CONFIG_FILE="extras/centOS/7/mock/epel-7-ppc64le.cfg"
 MAIN_CENTOS_REPO_RELEASE_URL="http://mirror.centos.org/altarch/7"
+
 
 # Fetch pull requests in case this job was triggered by one
 git clone $VERSIONS_REPO_URL $VERSIONS_REPO_DIR --no-checkout
@@ -31,6 +34,9 @@ python host_os.py \
 
 # creating the yum repository locally
 createrepo ./repository
+
+# Create BUILD_TIMESTAMP file with timestamp information
+echo "${TIMESTAMP}" > ./BUILD_TIMESTAMP
 
 # inform status to upload job
 touch SUCCESS
--- a/jenkins_jobs/trigger_nightly_host_os_build.groovy
+++ b/jenkins_jobs/trigger_nightly_host_os_build.groovy
@@ -46,6 +46,15 @@ job('trigger_nightly_host_os_build') {
 	}
       }
     }
+
+    copyArtifacts('build_host_os') {
+      buildSelector {
+        buildNumber('$BUILD_JOB_NUMBER')
+      }
+
+      includePatterns('BUILD_TIMESTAMP')
+    }
+
     shell(readFileFromWorkspace(
 	    'jenkins_jobs/trigger_nightly_host_os_build/post_build_script.sh'))
   }
--- a/jenkins_jobs/trigger_weekly_host_os_build.groovy
+++ b/jenkins_jobs/trigger_weekly_host_os_build.groovy
@@ -51,6 +51,15 @@ job('trigger_weekly_host_os_build') {
 	}
       }
     }
+
+    copyArtifacts('build_host_os') {
+      buildSelector {
+        buildNumber('$BUILD_JOB_NUMBER')
+      }
+
+      includePatterns('BUILD_TIMESTAMP')
+    }
+
     shell(readFileFromWorkspace(
 	    'jenkins_jobs/trigger_weekly_host_os_build/post_build_script.sh'))
   }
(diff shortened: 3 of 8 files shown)


jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -1,32 +1,35 @@
-# Create status json
+# ISO 8601 date with  minute precision
+CURRENT_DATE=$(date --utc -Iminutes)
```
ghost: The upload job is not restricted to run concurrently and with 4 slaves we may (unlikely) have up to 4 build jobs finishing at the same time. This will get worse once we have multiple builds running in the same slave. I think minute precision is not enough, we'd need second or, to be absolutely safe, the full timestamp.

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -1,32 +1,35 @@
...
+BASE_URL="http://${UPLOAD_SERVER_HOST_NAME}${UPLOAD_SERVER_BUILDS_DIR}/${BUILD_JOB_NUMBER}"
```
ghost: I believe this should have `${CURRENT_DATE}` instead of `${BUILD_JOB_NUMBER}`.

ghost (changes requested):

jenkins_jobs/upload_build_artifacts/script.sh
ghost: Good point.

I think nanosecond precision will suffice.
It would be very unlikely there would be collisions this way.

jenkins_jobs/upload_build_artifacts/script.sh
ghost: Fixed

ghost (approved):

ghost:
We need to check with @sathnaga if the test scripts can be changed to accommodate those new paths.

ghost:
I left a message on Slack to him. I get back to this PR when I get a response.

lisibann (changes requested):
The jobs: trigger_weekly_host_os_build and trigger_nightly_host_os_build uses the build job number (post_build_script.sh) to update build dir and latest build. Both needs to be reviewed/modified.

jenkins_jobs/build_host_os/script.sh
```diff
@@ -32,5 +35,9 @@ python host_os.py \
 # creating the yum repository locally
 createrepo ./repository
 
+# Create META file with timestamp information
+touch ./META
```
ghost: There's no need to "touch" the file, it will be created in the next command anyway. The SUCCESS file you see below is always empty.

jenkins_jobs/build_host_os/script.sh
```diff
@@ -32,5 +35,9 @@ python host_os.py \
...
+echo "${TIMESTAMP}" > ./META
```
ghost: I'd prefer to name it TIMESTAMP or BUILD_TIMESTAMP. META may make one think there's more information in this file.

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -1,32 +1,34 @@
-# Create status json
+BUILD_TIMESTAMP=$(cat ./META)
```
ghost: The archived file needs to be copied in the upload job groovy script.

ghost (changes requested):

ghost (approved):

jenkins_jobs/upload_build_artifacts/script.sh
ghost: This should be `$(cat ./BUILD_TIMESTAMP)`

jenkins_jobs/trigger_weekly_host_os_build.groovy
```diff
@@ -51,6 +51,13 @@ job('trigger_weekly_host_os_build') {
 	}
       }
     }
+
+    // Get the META file which contains the generated build timestamp
```
ghost: I don't see a reason for this comment anymore.

jenkins_jobs/trigger_nightly_host_os_build.groovy
```diff
@@ -46,6 +46,13 @@ job('trigger_nightly_host_os_build') {
 	}
       }
     }
+
+    // Get the META file which contains the generated build timestamp
```
ghost: I don't see a reason for this comment anymore.

ghost (changes requested):

ghost (approved):

jenkins_jobs/trigger_nightly_host_os_build.groovy
```diff
@@ -46,6 +46,12 @@ job('trigger_nightly_host_os_build') {
 	}
       }
     }
+
+    copyArtifacts('build_host_os') {
+      buildSelector('$BUILD_JOB_NUMBER') {}
```
ghost: This should be:
```
buildSelector {
    buildNumber('$BUILD_JOB_NUMBER')
}
```

jenkins_jobs/trigger_weekly_host_os_build.groovy
```diff
@@ -51,6 +51,12 @@ job('trigger_weekly_host_os_build') {
 	}
       }
     }
+
+    copyArtifacts('build_host_os') {
+      buildSelector('$BUILD_JOB_NUMBER') {}
```
ghost: This should be:
```
buildSelector {
    buildNumber('$BUILD_JOB_NUMBER')
}
```

ghost (changes requested):

ghost (approved):

PR #19, opened 2017-01-26
Execute upgrade-versions on nightly builds

lisibann:
Nightly builds are currently generated with the master branch of builds and versions repository. It would be better if it executed the upgrade-versions command to get the latest commits, so the developers could install the updated RPMs before the weekly releases.

### DIFF
--- a/jenkins_jobs/trigger_nightly_host_os_build/pre_build_script.sh
+++ b/jenkins_jobs/trigger_nightly_host_os_build/pre_build_script.sh
@@ -1,2 +1,15 @@
 RELEASE_DATE=$(date +%Y-%m-%d)
+COMMIT_BRANCH="nightly-${RELEASE_DATE}"
+
+# Upgrade versions
+# sudo yum install rpmdevtools
+python host_os.py \
+   --verbose \
+   upgrade-versions \
+       --committer-name "$GITHUB_BOT_NAME" \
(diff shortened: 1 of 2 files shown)


jenkins_jobs/trigger_nightly_host_os_build/pre_build_script.sh
```diff
@@ -1,2 +1,15 @@
...
+
+# Upgrade versions
+# sudo yum install rpmdevtools
+python host_os.py \
+   --verbose \
+   upgrade-versions \
+       --committer-name "$GITHUB_BOT_NAME" \
+       --committer-email "$GITHUB_BOT_EMAIL" \
+       --push-repo-url "ssh://git@github.com/${GITHUB_BOT_USER_NAME}/versions.git" \
+       --push-repo-branch "$COMMIT_BRANCH"
```
ghost, line 12: Since we don't want to merge the nightly commits, I think it's better to modify the upgrade-versions command to allow its usage without the push-repo parameters. We could then use the local commit and avoid adding all those unnecessary parameters to this job. Those pushes would also create a lot of branches in the bot GitHub repo, making it hard to navigate.

lisibann: I've created a PR for upgrade-versions modification required. After reviews I'll update this one.

PR #20, opened 2017-01-26
Build Host OS ISO on weekly and nightly builds

ghost:


jenkins_jobs/upload_iso/script.sh
```diff
@@ -0,0 +1,20 @@
...
+
+rsync -e "ssh -i $HOME/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa" \
+      --verbose --compress --stats --times --chmod=a+rwx,g+rwx,o- \
+      *.iso \
+      "${ISO_DIR_RSYNC_URL}/"
+
+rsync -e "ssh -i $HOME/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa" \
+      --verbose --compress --recursive --stats --times --chmod=a+rwx,g+rwx,o- \
+      *-CHECKSUM \
+      "${ISO_DIR_RSYNC_URL}/"
```
mopsfelder: What about defining an alias for rsync command with common parameters, for example:

```shell
alias rsync="rsync -e 'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
    --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"

rsync iso ${BUILD_DIR_RSYNC_URL}  # push empty directory
rsync *.iso ${ISO_DIR_RSYNC_URL}  # push iso image
rsync *-CHECKSUM ${ISO_DIR_RSYNC_URL}  # push checksum file
```

I think you do not need to add trailing `/` in the last two rsync commands.

ghost: That looks better indeed. I'm also applying it to the other upload job in a separate commit.

ghost (commented):
Wouldn't it be better to create a differently named alias?

This will be shadowing the rsync command throughout the script.

I believe something like `rsync_upload` would be better.

ghost:
@snowcrshd Yeah, that might be less confusing. Applying it.

ghost (approved):

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -1,34 +1,30 @@
...
+alias rsync_upload="rsync -e \
+    'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
+    --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"
 
-rsync -e "ssh -i $HOME/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa" \
-      --verbose --compress --recursive --stats --times --chmod=a+rwx,g+rwx,o- \
-      "repository" \
-      "${UPLOAD_SERVER_USER_NAME}@${UPLOAD_SERVER_HOST_NAME}:${UPLOAD_SERVER_BUILDS_DIR}/${BUILD_TIMESTAMP}/"
+rsync_upload STATUS ${BUILD_DIR_RSYNC_URL}/
+rsync_upload --recursive repository ${BUILD_DIR_RSYNC_URL}/
```
mopsfelder: I think you don't need the trailing `/` in these two commands.

ghost: It's needed in the first command to create the build directory. I am removing it and making the directory creation explicit to avoid confusion.

PR #23, opened 2017-02-03
Execute upgrade-versions on nightly builds

ghost:
It is more useful to have nightly builds that are always up-to-date with the top of the repositories' branches.


ghost:
This depends on https://github.com/open-power-host-os/builds/pull/174
And replaces https://github.com/open-power-host-os/infrastructure/pull/19

PR #24, opened 2017-02-06
Fix buggy build timestamp introduced in 555100fe5d9626a4ccc22b3feeaf4…

ghost:
…54bf6a4b7fd

Using the previous timestamp as the directory name caused issues
with YUM. Apparently it has troubles with unencoded URLs.

### DIFF
--- a/jenkins_jobs/build_host_os/script.sh
+++ b/jenkins_jobs/build_host_os/script.sh
@@ -1,5 +1,5 @@
 # ISO 8601 date with nanoseconds precision
-TIMESTAMP=$(date --utc -Ins)
+TIMESTAMP=$(date --utc +'%Y-%m-%dT%H:%M:%S.%N')
 VERSIONS_REPO_DIR=$(basename $VERSIONS_REPO_URL .git)
 MOCK_CONFIG_FILE="extras/centOS/7/mock/epel-7-ppc64le.cfg"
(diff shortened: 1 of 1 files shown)


farosas (approved):

jenkins_jobs/build_host_os/script.sh
```diff
@@ -1,5 +1,5 @@
 # ISO 8601 date with nanoseconds precision
-TIMESTAMP=$(date --utc -Ins)
+TIMESTAMP=$(date --utc +'%Y-%m-%dT%H:%M:%S:%N+%Z')
```
ghost: Do you know if yum has a problem with the `,`, the `+` or both? If it accepts `.`, we could use `+%Y-%m-%dT%H:%M:%S.%N+%Z` to keep following the standard. It would also be more close to Python's isoformat (e.g. `2017-02-06T17:05:19.278788`)

ghost: The problem is with the `,`. For some reason YUM translated it into an encoded space (i.e., `%20`). Not sure how it handles '.', though.

mopsfelder: I tested with `2017-02-02T11:43:17.134075170+0000` and it worked just fine from yum perspective.

ghost: Well, if it works, I'll update the PR. I'll drop the timezone too.

ghost (approved):

PR #25, opened 2017-02-06
Update Jenkins setup doc

alanoe:


farosas:
I think the documentation should not mix infrastructure setup with HostOS build/release process. Looking at the repo I can see that this could easily be used by people outside of HostOS builds to setup a Jenkins instance with GitHub integration but using another set of jobs and performing different tasks. Ideally I'd like to see:

- the first block of "Infrastructure description" replaced with a top-level description of what the infrastructure looks like (jenkins master + nodes + github integration)

- removal of mentions to jobs, hostos, release process, etc

- a separate section saying something like: "for HostOS, *this* is the ansible role, *these* are the jobs, *this* is the behavior and you should have *these* forks cloned. For another project replace/edit the groovy files and create a new ansible role"

What do you think?

alanoe:
Our entire infrastructure setup code makes assumptions based on the fact that it will be used for OpenPOWER Host OS builds. If it is a requirement to support a generic Jenkins + GitHub integration setup, IMHO code would have to change too so we are consistent. IMHO this has low priority and should be done in the future, in case the team confirms it is a requirement.

README.md
```diff
@@ -1,24 +1,28 @@
...
-"versions" repositories and generating weekly builds. The single components are
-independent, so you could execute just the parts that interest you. For example,
-you could execute only the Ansible playbook that prepares a host for building
-Host OS or just add a single job to a pre-existing Jenkins instance using its
-Groovy script.
+They will prepare a Jenkins instance capable of:
+- manually start a build
+- automatically start a build when there are pull requests to the "builds" and
+"versions" git repositories in GitHub
+- create weekly builds, which must have some extra artifacts
```
ghost: and nightly

README.md
```diff
@@ -1,24 +1,28 @@
...
+
+A Jenkins instance is composed of at least two nodes: a master one which
```
ghost: We're talking about our particular case (Jenkins could have only a master node with no slaves), so I'd write `The Jenkins instance` instead of `A Jenkins instance`.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+packages' versions in the "versions" git repository, create a build with all
+packages available and create release notes, by executing the corresponding
+commands from the build scripts in the "builds" git repository. Commits will
+be created for the local "versions" and "open-power-host-os.github.io" git
+repositories and pushed to the designated GitHub user's git repository. Ideally,
+this should be a "bot" user. Pull requests for those commits can be created
+manually and merged into the main organization/user git repository, if desired.
+
+Note: The default is to execute those periodic builds once a week (every Wednesday
+at 11 AM UTC-03:00).
```
ghost: The weekly builds execute at 11AM in the system's time. I've manually configured our machines to have Brazil's time (BRT/BRST), since our whole team is here. Jenkins gets this configuration automatically. Also note we're currently in UTC-2 due to summer time.

ghost: We could add a paragraph mentioning nightly builds:

There's another periodic job similar to the one above, which differs by executing optional checks during the packages builds and by not pushing commits to remote git repositories. It allows developers of the packages to have their new commits tested and to have a build with those commits as soon as possible.

Note: The default is to execute those periodic builds nightly (every day at 22 PM).

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+ - Create builds jobs
+
 They are described in detail below.
 
-### Prepare CentOS system
+### Prepare CentOS system(s)
+
+First, you will need to prepare at least one CentOS system. If the Jenkins
+instance will be used for testing, one system is enough; production instances
+should use more systems.  Currently, OpenPOWER Host OS has 1 Jenkins slave for
```
ghost: Extra space before `Currently`

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+validating the code from GitHub pull requests and 3 Jenkins slaves for builds.
```
ghost: This is not true. We have 4 slaves used for both builds and code validation. I also don't think this should be in documentation. We may change it at any time and is not a requirement to a particular setup.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+
+KVM virtual machines are recommended, but you could optionally use bare-metal or
+containers instead. If KVM virtualization is used, for each required system,
+you will need to create a KVM guest and install a CentOS ISO in it. It is
+recommended to use virt-install to do it.
```
ghost: There are other ways to do this instead of using virt-install that may be better for a particular setup, I don't think this phrase aggregates much.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+
+If you do not have a direct network connection to the system(s), you may
+setup a SSH proxy in ~/.ssh/config in your workstation to make the SSH
```
ghost: "an SSH", it starts with a vowel sound.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+proxy transparent. Example:
 
-First, you'll need to prepare at least one CentOS system. Virtual machines are
-recommended, but they could also be bare-metal or containers.
+Host ophostos-builds
```
ghost: I'd write `open-power-host-os-builds` to be consistent with the name in other places.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
-### Execute Ansible playbooks
+### Setup SSH credentials used to access upload server and GitHub from Jenkins slaves
 
-There are Ansible playbooks that set up the Jenkins master node and two Jenkins
-slave nodes, one for Host OS builds and one for validating the code from GitHub
-pull requests. There could be multiple slave nodes set up for parallel builds.
-The different roles (master, builds slave and validation slave) could even be
-set up in the same system, as desired.
+To automate SSH authentication between 2 hosts, it is necessary to:
+ - create a SSH key pair if you still do not have one
```
ghost: "an SSH"

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+ - add the target host to ~/.ssh/known_hosts in the source host. It is also
+   possible to do it by executing a simple SSH command from the source host to
+   the target host. Alternatively, you could disable "StrictHostKeyChecking"
+   for all SSH connections in the source host, but that may be a security hazard.
+ - add the SSH private key to ~/.ssh/ in the source host
```
ghost, line 101: This step is actually done by the ansible playbook, but I see no harm in leaving it here too.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+
+Note: The Jenkins playbooks may fail due to network errors. If you see HTTP
+request errors, try executing them again.
+
 
-### Manual Jenkins steps
+### Manually setup Jenkins via web UI
+
+If you do not have a direct network connection to the system(s), you may
+create a SSH tunnel in your workstation to be able to access the Jenkins
```
ghost: "an SSH"

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+web UI, which is accessible via HTTPS. Example:
+
+`sudo ssh -o "ProxyCommand ssh <user>@<proxy_server> -W %h:%p" -NL 443:localhost:443 root@<jenkins_master>`
```
ghost: I use a different command and I am not sure what `<proxy_server>` should be filled with. Maybe use a different name or explain in the text.

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
-configured, execute the seed job with a glob that matches only the
-administrative jobs (e.g. "jenkins_jobs/create_*.groovy"). This will prevent
-"trigger" jobs from being created, since they would fail if executed
-automatically without properly configured credentials.
+The Credentials plugin allows you to store and manage credentials in Jenkins.
+In the case of SSH credentials, the SSH key pair must have been created beforehand.
+The following credentials must be created in Jenkins:
 
-#### Create slaves credentials
+##### Setup SSH credentials used to access Jenkins slaves from Jenkins master in Jenkins
```
ghost: There are too many "Jenkins" in this phrase. I suggest: "Setup SSH credentials used to access Jenkins slaves from master"

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
 
-Create the credentials necessary to access the slave nodes:
-`https://myjenkins.com/job/create_ssh_credentials`
+Setup the SSH credentials necessary to access the slaves in Jenkins:
+`https://<jenkins_server>/job/create_ssh_credentials`
 
-#### Create GitHub credentials
+##### Setup credentials used to access GitHub API from Jenkins slave in Jenkins
```
ghost: I'd remove "in Jenkins"

README.md
```diff
@@ -27,115 +31,164 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+https://github.com/blog/1509-personal-api-tokens for more information on
 how to create those tokens.
 
 Next, go to Jenkins home -> "Manage Jenkins" -> "Configure System" and look
-for "GitHub Pull Request Builder". Select the credentials you've created and
-set a meaningful description. The other fields do not need to be modified. You
-can use the "Test Credentials" button to make sure the permissions to your
-repository are correct. You'll need at least push and pull permissions.
+for "GitHub Pull Request Builder". Select the credentials you have created and
+set a meaningful description. The other job parameters values do not need to be
```
ghost: This is not a job. Either "fields" or "parameters values" are fine.

ansible/README.md
```diff
@@ -1,11 +1,32 @@
...
 
 ## Running Ansible playbooks
 
-You can run Ansible playbooks by executing `ansible-playbook` command:
+First, update the host in the hosts.ini section which corresponds to the
+playbook which will be executed (eg if you will execute the jenkins-master
+playbook, update the host in [jenkins-master] section in hosts.ini)
+
+Then, update the variables values in vars.yaml. If you are creating a production
+Jenkins server, provide the production GitHub organization (open-power-host-os) in
```
ghost: Other organization may replicate the setup, and they will not want to use our organization's name (open-power-host-os).

ansible/README.md
```diff
@@ -1,11 +1,32 @@
...
+github_organization_name and ghprob_admin_organization variables. If you are creating
```
ghost: `ghprb` instead of `ghprob`

ansible/README.md
```diff
@@ -1,11 +1,32 @@
...
+a test Jenkins server, provide your GitHub organization and user name in
+github_organization_name, ghprb_admin_user and ghprb_admin_organization
+variables.
```
ghost: This looks like the user should use the GitHub organization name in the `github_organization_name` variable, which is not true for our custom setups. I suggest:
"you may provide your user name in github_organization_name and ghprb_admin_user variables and leave ghprb_admin_organization empty."

ansible/README.md
```diff
@@ -1,11 +1,32 @@
...
+
+TODO: what should be the value of ghprb_admin_user in a production Jenkins server?
```
ghost: It should be empty, which is the default value. I'd avoid leaving TODOs in the external documentation.

ansible/README.md
```diff
@@ -1,11 +1,32 @@
...
+repository and run:
+
+`yum install -y ansible`
 
 ## Running Ansible playbooks
 
-You can run Ansible playbooks by executing `ansible-playbook` command:
+First, update the host in the hosts.ini section which corresponds to the
+playbook which will be executed (eg if you will execute the jenkins-master
+playbook, update the host in [jenkins-master] section in hosts.ini)
```
ghost: One may understand that only the host(s) that will be configured should be filled at execution time. I'd say something like: "update all the hosts in the hosts.ini file, paying attention to the groups they belong to. When executing a playbook, only the hosts in the corresponding group will be updated."

ghost (changes requested):

README.md
alanoe: Done

README.md
alanoe: Done

README.md
alanoe: OH> The weekly builds execute at 11AM in the system's time. I've manually configured our machines to have Brazil's time (BRT/BRST), since our whole team is here. Jenkins gets this configuration automatically. 

IMHO we should document the whole setup; if production Jenkins instance breaks tomorrow, I should have all the steps required to replicate it in a new system documented. Therefore, I should add the manual time setup to the doc. This text may remain as it is.

I see a trend of creating a generic Jenkins setup doc. I think that is an optional plus and that  doc should be kept in a separate place (an alternative is to do the opposite: keep the generic doc here and create a specific one internally). The specific doc should be only about replicating our current setup in a new system, mixing it with alternatives make the doc harder to follow. 

Also, I think that our infra setup code was not designed to work as a generic Jenkins setup and have specificities that only our team will use and we do not teach how to convert this to a generic setup. Therefore, any person interested in a generic setup needs to get the code and work by himself/herself. I don't like this approach. Another point is that IMHO this use case may not even exist (for instance, there may be better generic Jenkins setups available in the web or this project is too hidden of the greater public so a developer working in a very distinct project will never look for a generic setup here) ; as I mentioned in our last meeting, we waste time trying to aim a public which may not even exist or cover options which may never be used. I am not against working in non-required use cases, but at least we should be certain a lot of people will take advantage in using what we worked on.

README.md
alanoe: Done

README.md
alanoe: IMHO we should. If our current infrastructure fails, we should have *all* steps to replicate it in another place documented so *anyone* can do it. We should not depend on knowledge which is in the minds of current members of our team.

README.md
alanoe: I don't care about alternate setups, I care about *our* setup. Why you assume the reader knows how to create VMs? That's why I had all the virt-install commands documented in PowerKVM infra setup doc. I prefer to not make any assumptions about reader prior knowledge.

This is related to the generic setup x our particular setup doc discussion I had above.

README.md
alanoe: this rule is tricky. Done.

README.md
alanoe: Done

README.md
alanoe: Done

README.md
alanoe: Done

README.md
alanoe: proxy server is the server which will be the intermediate (proxy) between you and the Jenkins server. I'll make it explicit in the doc.

README.md
alanoe: Done

README.md
alanoe: Done

ansible/README.md
alanoe: I want all steps required to replicate *our* infrastructure documented. This is related to the generic x specific setup discussion I started above.

README.md
alanoe: Done

ansible/README.md
alanoe: Done

ansible/README.md
alanoe: I agree. Done

README.md
ghost: Having a setup that works for other teams is not an optional plus, it is a goal that we must work towards.
This can't be done all at once, so the first step was to make this work for us. The future steps will be to make this easy for everybody else.
The specific configuration for our use case should be in an internal document (it does not exist yet) and not mixed here to avoid confusing whoever comes to actually replicate this setup. At least, they should be able to differentiate what's optional and what's not.

README.md
ghost: This is information that may change for us over time and may not make sense for other system configurations. It should be kept internally only.

ansible/README.md
ghost: Instructions that do not work for other teams should be kept internally.

README.md
alanoe: OH> Having a setup that works for other teams is not an optional plus, it is a goal that we must work towards.

Why? Why other teams need to submit pull requests to a git hub organization and have builds triggered automatically? Why other teams need to create release notes? Why other teams need to update versions in packages yamls based in a set of git repositories if we are the only ones which maintain multiple packages in OP Host ? The only feature I really see other teams need is to build OP Host OS (packages and ISO) and they don't even need Jenkins to do this, it can be done much more easily in the command line! Having a graphical UI is a benefit, but is it a must? Do we know of at least ONE team which uses this Jenkins infrastructure or wants to?

OH> This can't be done all at once, so the first step was to make this work for us. The future steps will be to make this easy for everybody else.

If the first step is to make it work for us (and this makes sense), why are we spending time *now* in a doc which is not for us? This is not consistent. Why not build a doc which is for us *first* and later work on a generic doc which would work for others? 

OH> The specific configuration for our use case should be in an internal document (it does not exist yet) 

So we are sacrificing having a full documentation about how to replicate *our* infrastructure in case of disaster to support a documentation which is not used by anyone currently?

README.md
alanoe: I agree it could be kept internally, I'm just saying that should be our doc priority ONE: document all steps to replicate *our* infrastructure. If you agree, I'll change priorities here and work on documenting these steps internally.

README.md
alanoe: I've mentioned that the Ansible playbooks do it.

README.md
```diff
@@ -27,115 +31,167 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+repositories and pushed to the designated GitHub user's git repository. Ideally,
+this should be a "bot" user. Pull requests for those commits can be created
+manually and merged into the main organization/user git repository, if desired.
+The default is to execute those periodic builds once a week (every Wednesday
+at 11 AM, Jenkins master's timezone).
+
+There's another periodic job similar to the one above, which differs by executing
+additional testing during the packages builds and by not pushing commits to remote
+git repositories. It allows developers of the packages to have their new commits
+tested and to have a build with those commits as soon as possible. Note: The default
```
ghost: Remove the "Note:", like you did above.

README.md
```diff
@@ -27,115 +31,167 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+is to execute those periodic builds nightly (daily at 22 PM, Jenkins master's timezone).
```
ghost: "nightly (daily" looks kind of weird. Maybe "(every day" looks better?

README.md
```diff
@@ -27,115 +31,167 @@ repositories' pull requests by using the GitHub Pull Request Builder plugin. It
...
+proxy transparent. Example:
+
+Host open-power-hostos-builds
+    HostName <target_server>
+    ProxyCommand ssh <myuser>@<proxy_server> nc <target_server> 22
 
-First, you'll need to prepare at least one CentOS system. Virtual machines are
-recommended, but they could also be bare-metal or containers.
+After the SSH proxy is configured, simply use the following to connect to the system:
+`ssh <myuser>@ophost-builds`
```
ghost: `<myuser>@open-power-hostos-builds`

ghost (changes requested):
Just one correction and two minor comments.

README.md
alanoe: Missed it. Done

README.md
alanoe: Daily means "every day", not "during the day", and refers to a periodicity as weekly, monthly and yearly do. Nightly does not refer to periodicity, but only to the period of the day when it happens. IMHO daily is more accurate than nightly here.

README.md
alanoe: Done

ghost (approved):

farosas (approved):

PR #26, opened 2017-02-06
Add rpmlint and yamllint validation jenkins jobs

alanoe:


jenkins_jobs/trigger_yamllint_from_versions_repo/script.sh
```diff
@@ -0,0 +1,2 @@
+YAML_FILES=$(find . -name "*.yaml")
+yamllint $YAML_FILES
```
ghost: Shouldn't we be executing the validate_yamls.py script here? This will require cloning the builds repo.
I've also tested and this command failed because it's not reading the .yamllint file.
The same applies for rpmlint.

ghost (changes requested):

jenkins_jobs/trigger_yamllint_from_versions_repo/script.sh
alanoe: Done

jenkins_jobs/trigger_rpmlint_from_versions_repo.groovy
```diff
@@ -1,19 +1,20 @@
-job('trigger_rpmlint_from_builds_repo') {
+job('trigger_rpmlint_from_versions_repo') {
   label('validation_slave_label')
   concurrentBuild()
   parameters {
-    stringParam('sha1', '', 'SHA-1 of the commit to validate.')
+    stringParam('builds_sha1', 'master', 'SHA-1 of the builds commit')
```
farosas: Is this variable here so that we can run the job manually? Because I don't think the trigger plugin has a way of setting it, right?

jenkins_jobs/trigger_rpmlint_from_versions_repo.groovy
```diff
@@ -1,19 +1,20 @@
-job('trigger_rpmlint_from_builds_repo') {
```
farosas: It seems to me that all changes to this file should be in the other commit.

jenkins_jobs/trigger_yamllint_from_versions_repo.groovy
```diff
@@ -0,0 +1,43 @@
...
+    stringParam('builds_sha1', 'master', 'SHA-1 of the builds commit.')
+    stringParam('sha1', 'master', 'SHA-1 of the commit to validate.')
+  }
+  properties {
+    githubProjectUrl("https://github.com/${GITHUB_ORGANIZATION_NAME}/versions/")
+  }
+  scm {
+    git {
+      remote {
+	url("https://github.com/${GITHUB_ORGANIZATION_NAME}/versions.git")
```
farosas: builds, isn't it? You are checking out `builds_sha1`.

farosas (changes requested):

jenkins_jobs/trigger_yamllint_from_versions_repo.groovy
alanoe: Done

farosas (approved):

jenkins_jobs/trigger_rpmlint_from_versions_repo.groovy
```diff
@@ -0,0 +1,51 @@
...
+		'Commit ID to checkout from the versions repository.')
+  }
+  properties {
+    githubProjectUrl("https://github.com/${GITHUB_ORGANIZATION_NAME}/versions/")
+  }
+  scm {
+    git {
+      remote {
+	url('$BUILDS_REPO_URL')
+	refspec('+refs/pull/*:refs/remotes/origin/pr/*')
```
ghost: This is not needed since this is triggered from the versions repo only.
Also, is the `heads` spec still fetched when you use this? See https://github.com/open-power-host-os/infrastructure/blob/master/jenkins_jobs/build_host_os.groovy#L31

jenkins_jobs/trigger_yamllint_from_versions_repo.groovy
```diff
@@ -0,0 +1,43 @@
...
+	refspec('+refs/pull/*:refs/remotes/origin/pr/*')
```
ghost: Same comment as in RPMlint.

jenkins_jobs/trigger_yamllint_from_versions_repo.groovy
```diff
@@ -0,0 +1,43 @@
+job('trigger_yamllint_from_versions_repo') {
+  label('validation_slave_label')
+  concurrentBuild()
+  parameters {
+    stringParam('builds_sha1', 'master', 'SHA-1 of the builds commit.')
+    stringParam('sha1', 'master', 'SHA-1 of the commit to validate.')
```
ghost: Please be consistent across jobs. If you are allowing manual execution for RPMlint, do ti here too.

ghost (changes requested):
Could you do a rebase on top of main repo's master to keep the history cleaner?

jenkins_jobs/trigger_yamllint_from_versions_repo.groovy
```diff
@@ -0,0 +1,51 @@
...
+		'Commit ID to checkout from the versions repository.')
+  }
+  properties {
+    githubProjectUrl("https://github.com/${GITHUB_ORGANIZATION_NAME}/versions/")
+  }
+  scm {
+    git {
+      remote {
+	    url('$BUILDS_REPO_URL')
+	    refspec('+refs/heads/*:refs/remotes/origin/* ')
```
ghost: This is not needed, since it's the default.

ghost (approved):

PR #28, opened 2017-02-09
Add job to update versions README file

ghost:
I'll update this PR with a commit creating a trigger for the GitHub Merge event, in order to
start the update job automatically.

### DIFF
--- a/jenkins_jobs/update_versions_readme/script.sh
+++ b/jenkins_jobs/update_versions_readme/script.sh
@@ -1,0 +1,9 @@
+python host-os.py \
+  --verbose \
+  update-versions-readme \
+    --build-versions-repository-url ${VERSIONS_REPO_URL} \
+    --build-version ${VERSIONS_REPO_COMMIT} \
+    --updater-name "${GITHUB_BOT_NAME}" \
+    --updater-email "${GITHUB_BOT_EMAIL}" \
+    --push-repo-url "ssh://git@github.com/${GITHUB_BOT_USER_NAME}/versions.git" \
+    --push-repo-branch "${COMMIT_BRANCH}"
--- a/jenkins_jobs/update_versions_readme.groovy
+++ b/jenkins_jobs/update_versions_readme.groovy
@@ -1,0 +1,48 @@
+job('update_versions_readme') {
+  label('!master')
+
+  parameters {
+    stringParam('BUILDS_REPO_URL',
+		"https://github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
+		'URL of the builds repository.')
+
+    stringParam('BUILDS_REPO_COMMIT', 'master',
+		'Commit ID to checkout from the builds repository.')
+
+    stringParam('VERSIONS_REPO_URL',
+		"https://github.com/${GITHUB_ORGANIZATION_NAME}/versions.git",
+		'URL of the versions repository.')
+
+    stringParam('VERSIONS_REPO_COMMIT', 'master',
+		'Commit ID to checkout from the versions repository.')
+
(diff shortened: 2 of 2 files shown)


jenkins_jobs/update_versions_readme/script.sh
```diff
@@ -0,0 +1,9 @@
+python host-os.py \
+  --verbose \
+  update-versions-readme \
+    --build-versions-repository-url ${VERSIONS_REPO_URL} \
+    --build-version ${VERSIONS_REPO_COMMIT} \
+    --updater-name "${GITHUB_BOT_NAME}" \
+    --updater-email "${GITHUB_BOT_EMAIL}" \
+    --push-repo-url "ssh://git@github.com/${GITHUB_BOT_USER_NAME}/versions.git" \
```
alanoe, line 8: why you hardcode git@github.com and versions.git here and not in build-versions-repository-url value? Please be consistent

jenkins_jobs/update_versions_readme/script.sh
```diff
@@ -0,0 +1,9 @@
+python host-os.py \
+  --verbose \
+  update-versions-readme \
```
alanoe, line 3: Why create a update-versions-readme subcommand instead of doing this in existing update-versions subcommand?

alanoe (changes requested):

jenkins_jobs/update_versions_readme/script.sh
ghost: Because this is the way it is done in other scripts.

Refer to https://github.com/open-power-host-os/infrastructure/blob/master/jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh, for instance.

I'm being consistent with other scripts.

jenkins_jobs/update_versions_readme/script.sh
ghost: Because this is more flexible. `upgrade-versions` is (currently) only run when we want to _upgrade_ the versions of a few selected packages.

By having a separate update-versions, we are not tied to the use cases of upgrade-versions. This allows us to update the README whenever a package that is not targeted by the `upgrade-versions` script is changed.

alanoe: Your point makes sense, but can't we update the versions of *all* packages in the README and update the versions of a sublist of packages in yamls in upgrade-versions? IMHO it is better to automatically update README whenever the yamls are updated to avoid extra work and accidental inconsistencies.

jenkins_jobs/update_versions_readme/script.sh
alanoe: I see. I'll write a patch to make this consistent in all scripts then.

jenkins_jobs/update_versions_readme/script.sh
ghost: > IMHO it is better to automatically update README whenever the yamls are updated to avoid extra work and accidental inconsistencies.

That's the idea.
Whenever there is a Merge event on the `versions` repository, we trigger an `update-versions` job. Hence my comment on updating this PR with a `trigger_*` job.

See my first comment on this PR:

> I'll update this PR with a commit creating a trigger for the GitHub Merge event, in order to
> start the update job automatically.

jenkins_jobs/update_versions_readme/script.sh
ghost: Cool.

ghost:
This depends on https://github.com/open-power-host-os/builds/pull/181

alanoe:
We agreed we do not need a Jenkins job for this, therefore I'll close this PR

PR #29, opened 2017-02-09
Allow weekly trigger to automatically open a GitHub pull request

farosas:


scripts/github_api
```diff
@@ -0,0 +1,106 @@
+#!/bin/python
```
ghost: Since we don't who might use this, better specify `python2`

scripts/github_api
```diff
@@ -0,0 +1,106 @@
...
+request = None
+
+
+def _request(auth, method, endpoint, **kwargs):
+    methods = {
+        'POST': requests.post
+    }
+
+    r = methods[method](endpoint, auth=auth, **kwargs)
+    response = json.loads(r.text)
```
ghost: This is only used inside the `except` block, it could be there.

scripts/github_api
```diff
@@ -0,0 +1,106 @@
...
+                          '<repo owner>/<repo>:<branch>'))),
+                (('--body',), dict(
+                    help='pull request body text')),
+            ],
+        },
+    }
+
+    parser = argparse.ArgumentParser(description='Wrapper to GitHub API')
+    parser.add_argument('user', help='github user name')
+    parser.add_argument('passwd', help='github password or token')
```
ghost: I think we can afford two more letters in the variable name :stuck_out_tongue:

scripts/github_api
```diff
@@ -0,0 +1,106 @@
...
+        "body": body,
+        "head": src,
+        "base": base_branch,
+    }
+
+    request('POST', endpoint, data=json.dumps(payload))
+
+
+if __name__ == '__main__':
+    commands = {
```
ghost: This is a constant. Uppercase.

scripts/github_api
```diff
@@ -0,0 +1,106 @@
...
+    parser.add_argument('user', help='github user name')
+    parser.add_argument('passwd', help='github password or token')
+
+    subparsers = parser.add_subparsers()
+
+    for cmd_name, cmd in commands.items():
+        cmd_parser = subparsers.add_parser(
+            cmd_name, description=cmd['description'])
+
+        for arg_name, arg_opts in cmd['arguments']:
```
ghost: This is a tuple, so I think `arg_names` would be more appropriate.

ansible/roles/github-api-wrapper/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - setup
+
+- name: Install GitHub API wrapper
+  copy:
+    src={{item.src}} dest={{item.dest}}
+    owner={{item.owner}} group={{item.group}} mode={{item.mode}}
+  with_items:
+    - { src: "{{git_repo_root_dir.stdout}}/scripts/github_api",
+        dest: "/usr/local/bin",
+        owner: jenkins, group: jenkins, mode: "0500" }
```
ghost: Why don't we let other users execute it by setting `0555`?

jenkins_jobs/trigger_weekly_host_os_build.groovy
```diff
@@ -86,5 +86,8 @@ job('trigger_weekly_host_os_build') {
   wrappers {
     timestamps()
     preBuildCleanup()
+    credentialsBinding {
+      usernamePassword('GITHUB_USERNAME', 'GITHUB_PASSWORD', 'github-user-pass-credentials')
```
ghost: `GITHUB_USER_NAME`

scripts/github_api
```diff
@@ -0,0 +1,114 @@
...
+import re
+import requests
+import sys
+
+CLIENT_ERROR = 422
+API_BASE_URL = "https://api.github.com/"
+
+
+# see _setup_auth function below
+request = None
```
ghost: We could define this function by raising an exception that informs `setup_auth` must be called first.

scripts/github_api
farosas: It's used in line 39

ghost (approved):

scripts/github_api
```diff
@@ -0,0 +1,118 @@
...
+CLIENT_ERROR = 422
+API_BASE_URL = "https://api.github.com/"
+
+
+def request(*args, **kwargs):
+    raise Exception("The '_setup_auth' function must be called once "
+                    "before a request can be made")
+
+
+def _setup_auth(user, password):
```
alanoe, line 19: Add documentation to all methods

scripts/github_api
```diff
@@ -0,0 +1,118 @@
...
+    auth = (user, password)
+    global request
+    request = partial(_request, auth)
+
+
+def _request(auth, method, endpoint, **kwargs):
```
alanoe: I have no idea what will be done with these keyword arguments. HTTP headers? Maybe method doc will make it clearer

scripts/github_api
```diff
@@ -0,0 +1,118 @@
...
+class PullRequest:
+    """
+    https://developer.github.com/v3/pulls/
+    """
+    def __init__(self, repo):
+        self.repo = repo
+        self._endpoint = API_BASE_URL + "repos/%(repo)s/pulls"
+
+    def open(self, title, body, src_branch, dst_branch):
+        endpoint = self._endpoint % dict(repo=self.repo)
```
alanoe: minor: I'd prefer to avoid HTTP library specific terminology, such as "endpoint" and "payload" and use the names defined by the HTTP protocol: URL and body.

alanoe: Why not do this in __init__() if repo is never changed after it?

scripts/github_api
```diff
@@ -0,0 +1,118 @@
...
+    # program and command names which are not known by the subparser
+    cmd_args = vars(cmd_parser.parse_args(sys.argv[4:]))
+
+    _setup_auth(args.user, args.password)
+
+    for key, value in cmd_args.items():
+        if value is None:
+            cmd_args.pop(key)
+
+    commands[args.command]['function'](**cmd_args)
```
alanoe, line 171: I'd prefer to split libraries and executables in separate files, as we probably will want to call this GitHub API from our code in the future, but up to you

scripts/github_api
```diff
@@ -0,0 +1,118 @@
...
+            "head": src_branch,
+            "base": dst_branch,
+        }
+
+        self._set_pr_info(request('POST', endpoint, data=json.dumps(payload)))
+
+    def _set_pr_info(self, pr_info):
+        for k, v in pr_info.items():
+            if type(v) != dict:
+                setattr(self, k, v)
```
alanoe: I am not sure what you are doing here. pr_info is the HTTP request response, right? (1) I'd prefer to mention "response" somewhere so that is clearer. (2) The response body from GitHub API is not always a JSON? In that case, you could use json()

ansible/roles/github-api-wrapper/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - setup
+
+- name: Install GitHub API wrapper
+  copy:
+    src={{item.src}} dest={{item.dest}}
+    owner={{item.owner}} group={{item.group}} mode={{item.mode}}
+  with_items:
+    - { src: "{{git_repo_root_dir.stdout}}/scripts/github_api",
+        dest: "/usr/local/bin",
+        owner: jenkins, group: jenkins, mode: "0555" }
```
alanoe, line 17: Why the owner has no write access (555 permission)?

ansible/roles/jenkins-master/defaults/main.yaml
```diff
@@ -36,3 +36,5 @@ jenkins_plugins:
   - timestamper
```
alanoe, line 36: Regarding commit longer msg, we do not *need* this. I suggest instead:

=> We need this to allow the GitHub API wrapper to communicate with GitHub API without duplicating the credentials in Jenkins and in the filesystem.

Side comment: I'd prefer to use only this GitHub API wrapper instead of doing half of things using it and another half using GitHub Pull Request Builder Jenkins plugin, but the opinion of the rest of the team may differ.

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
```diff
@@ -2,6 +2,17 @@ VERSIONS_REPO_URL="https://github.com/${GITHUB_ORGANIZATION_NAME}/versions.git"
 RELEASE_DATE=$(date +%Y-%m-%d)
 COMMIT_BRANCH="release-${RELEASE_DATE}"
 
+create_pr() {
```
alanoe: Please use create_pull_request()  , more intuitive

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
```diff
@@ -22,5 +35,7 @@ python host_os.py \
 	--push-repo-url "ssh://git@github.com/${GITHUB_BOT_USER_NAME}/${GITHUB_ORGANIZATION_NAME}.github.io.git" \
 	--push-repo-branch "$COMMIT_BRANCH"
 
+create_pr "${GITHUB_ORGANIZATION_NAME}.github.io"
+
 echo "VERSIONS_REPO_COMMIT=$COMMIT_BRANCH" > BUILD_PARAMETERS
 echo "$RELEASE_DATE" > WEEKLY_DIR_NAME
```
alanoe: side-comment not related to this patch: pre_build_script.sh should be entirely in shell script (instead of deferring to a Python script)? what is the current criteria to use Python or shell script in our automation code? In PowerKVM, we used Python for everything, shell script steps in Jenkins only called Python code.

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
```diff
@@ -2,6 +2,17 @@ VERSIONS_REPO_URL="https://github.com/${GITHUB_ORGANIZATION_NAME}/versions.git"
 RELEASE_DATE=$(date +%Y-%m-%d)
 COMMIT_BRANCH="release-${RELEASE_DATE}"
 
+create_pr() {
+    local dest_repo=$1
+
+    # the GITHUB_USER_NAME and GITHUB_PASSWORD variables below refer to
+    # the credentials owner, which is not necessarily the same as the
+    # source repo owner - GITHUB_BOT_NAME in this case
```
alanoe, line 10: The text makes it look like that the source repo owner is GITHUB_BOT_NAME, but I think that is the credentials owner, right? I suggest you rephrase it.

alanoe (changes requested):

scripts/github_api
farosas: That makes sense, but since we need this to be available for the jenkins jobs we would either have to clone the infra repo in every job or properly install the script via pip. I think we can keep it like this for now.

scripts/github_api
farosas: Right. This just sets the response values as attributes of the class.

ansible/roles/github-api-wrapper/tasks/main.yaml
farosas: Well, first of all I don't even see why another user on the Jenkins slave machine would want to execute this script, so I'm already hesitant of 555 instead of 500. Answering your question along the same lines, I don't see why (and how) the 'jenkins' user would want to write to this file.

ansible/roles/jenkins-master/defaults/main.yaml
farosas: Regarding the commit message, I don't get what you mean. Are you implying that we could keep an encrypted file in the Jenkins node and read/decrypt it when using it in the script?

I agree about using only this script, but the benefit of this will have to weighted against the added time and maintenance effort

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
farosas: No. That's what it means. GITHUB_USER_NAME and GITHUB_PASSWORD are the credential variables and GITHUB_BOT_NAME is the name of the source repo owner.

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
farosas: Well, since the .sh script is fairly simple, I would like to skip the extra level of indirection that a python script would create. It is much simpler for instance to understand that `echo "$RELEASE_DATE" > WEEKLY_DIR_NAME` creates a file than to read that from a script somewhere else. I think it makes it easier to manage and interact with the jobs if we have a little snippet of code to read. There's also the fact that we already have cmdline tools that do their work quite well (rsync, echo, date, bash parameter substitution) so I see no need to re-implement them in python. In general, I think we should keep what is related to Jenkins, jobs variable manipulation, job flow, inter-job communication (like BUILD_PARAMETERS) or things that can be accomplished with cmdline tools (like rsync) as shell script.

However, I do understand your point and that's why I wrote the github_api script instead of making a thread ball of curl and grep here.

ansible/roles/jenkins-master/defaults/main.yaml
alanoe: Regarding the commit msg, yes, we could do that. I prefer as you implemented, I'd just like to make it clearer in commit msg why we are doing this instead of putting the credentials in the system where the script would run (more obvious solution).

jenkins_jobs/trigger_weekly_host_os_build/pre_build_script.sh
alanoe: Thanks for clarifying.

ansible/roles/github-api-wrapper/tasks/main.yaml
alanoe: I was thinking about debugging, but we can write as root user anyway, so fine.

scripts/github_api
alanoe: +1

scripts/github_api
alanoe: Could you make the shebang consistent in all executable scripts (eg host_os.py) in a separate patch?

scripts/github_api
```diff
@@ -0,0 +1,171 @@
...
+            API_BASE_URL, "repos/%(repo)s/pulls" % dict(repo=self.repo))
+
+    def open(self, title, body, src_branch, dst_branch):
+        """
+        Opens a GitHub pull-request
+
+        Args:
+            title (str): title text
+            body (str): body text
+            src_branch (str): destination repository. Format:
```
alanoe: Is this documentation correct?

scripts/github_api
```diff
@@ -0,0 +1,171 @@
...
+        """
+        payload_body = {
+            "title": title,
+            "body": body,
+            "head": src_branch,
+            "base": dst_branch,
+        }
+
+        self._set_attributes(
+            request('POST', self._url, data=json.dumps(payload_body)))
```
alanoe, line 94: I'd prefer to put request() result in an intermediate 'response' or 'http_response' variable to make code clearer, but up to you

scripts/github_api
```diff
@@ -0,0 +1,171 @@
...
+
+
+class PullRequest:
+    """
+    https://developer.github.com/v3/pulls/
+    """
+    def __init__(self, repo):
+        """
+        Args:
+            repo (str): git repository path. Format <repo owner>/<repo>
```
alanoe: missing colon (:) after "Format"

alanoe (changes requested):

scripts/github_api
farosas: Nah, the fact that I'm having to stack pull-requests is introducing some noise. I'll fix and resend.

PR #31, opened 2017-02-14
Add trigger iso job

farosas:
This allows us to start a build_host_os_iso build with a comment in the pull-request. This job is a bit different from the other triggers because it waits until the corresponding build_host_os build finishes to get its build id and use it during its own build.

### DIFF
--- a/jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
+++ b/jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
@@ -23,6 +23,7 @@ job('trigger_host_os_build_from_builds_repo') {
       extensions {
 	commitStatus {
 	  context('Build Host OS')
+	  statusUrl('${JENKINS_URL}/job/build_host_os/${TRIGGERED_BUILD_NUMBER_build_host_os}')
 	}
       }
     }
--- a/jenkins_jobs/trigger_host_os_iso_build_from_versions_repo.groovy
+++ b/jenkins_jobs/trigger_host_os_iso_build_from_versions_repo.groovy
@@ -1,0 +1,57 @@
+job('trigger_host_os_iso_build_from_versions_repo') {
+  label('!master')
+  logRotator {
+    numToKeep(30)
+  }
+  concurrentBuild()
+  throttleConcurrentBuilds {
+    maxPerNode(1)
+  }
+  parameters {
+    stringParam('sha1', '', 'SHA-1 of the commit to build.')
+    stringParam('GITHUB_ORGANIZATION_NAME',
+		"${GITHUB_ORGANIZATION_NAME}",
+		'GitHub organization from where the Host OS repositories will be checked out.')
+  }
+  properties {
+    githubProjectUrl("https://github.com/${GITHUB_ORGANIZATION_NAME}/versions/")
+  }
+  triggers {
+    githubPullRequest {
+      userWhitelist("${GHPRB_ADMIN_USER}")
+      orgWhitelist("${GHPRB_ADMIN_ORGANIZATION}")
+      allowMembersOfWhitelistedOrgsAsAdmin()
+      cron('H/5 * * * *')
+      triggerPhrase('.*start\\W+iso.*')
+      onlyTriggerPhrase()
+      extensions {
+	commitStatus {
+	  context('Build Host OS ISO')
+	  statusUrl('${JENKINS_URL}/job/build_host_os_iso/${TRIGGERED_BUILD_NUMBER_build_host_os_iso}')
+	}
+      }
+    }
+  }
+  steps {
+    shell(readFileFromWorkspace(
+	    'jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh'))
+    downstreamParameterized {
+      trigger('build_host_os_iso') {
+	block {
+	  buildStepFailure('FAILURE')
+	  failure('FAILURE')
+	  unstable('UNSTABLE')
+	}
+	parameters {
+	  propertiesFile("BUILD_PARAMETERS", true)
+	}
+      }
+    }
+  }
+  wrappers {
+    timestamps()
+    credentialsBinding {
+      usernamePassword('GITHUB_USER_NAME', 'GITHUB_PASSWORD', 'github-user-pass-credentials')
+    }
+  }
+}
--- a/jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
+++ b/jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -e
+set -x
+
+
+LISTENED_CONTEXT="Build Host OS"
+
+
+get_build_state(){
+    # this will evaluate the queried keys as variables: state=<state>,
+    # target_url=<url>
+    eval $(github_api "$GITHUB_USER_NAME" "$GITHUB_PASSWORD" \
+                      query_status "${GITHUB_ORGANIZATION_NAME}/versions" \
+                      "$ghprbPullId" "$LISTENED_CONTEXT" --state \
+                      --target-url || echo "exit 1")
+}
+
+get_pr_state(){
+    # this will evaluate the queried keys as variables: state=<state>,
+    # title=<title>
+    eval $(github_api "$GITHUB_USER_NAME" "$GITHUB_PASSWORD" \
+                      query_pr "${GITHUB_ORGANIZATION_NAME}/versions" \
+                      "$ghprbPullId" --state --title || echo "exit 1")
+}
+
+state="pending"
+get_build_state
+while [ $state == "pending" ]; do
+    echo "Waiting for $LISTENED_CONTEXT to finish..."
+    echo "$target_url"
+    sleep 1m
+    get_build_state
+done
+
+if [ "$state" == "failure" -o "$state" == "error" ]; then
+    echo "$LISTENED_CONTEXT failed, aborting."
+    echo "$target_url."
+    exit 1
+fi
+
+if [ "$state" == "success" ]; then
+    echo "$LISTENED_CONTEXT succeeded."
+    echo "$target_url"
+    get_pr_state
+    if [ "$state" == "closed" ]; then
+        echo "Pull request ${ghprbPullId}: \"${title}\" was closed, aborting."
+        echo "$ghprbPullLink"
+        exit 1
+    fi
+fi
+
+echo "BUILD_JOB_NUMBER=$(basename $target_url)" > BUILD_PARAMETERS
+echo "Triggering ISO build from $target_url ..."
+
+exit 0
--- a/scripts/github_api
+++ b/scripts/github_api
@@ -42,7 +42,8 @@ def _request(auth, method, url, **kwargs):
         dict: HTTP response
     """
     methods = {
-        'POST': requests.post
+        'POST': requests.post,
+        'GET': requests.get,
     }
 
     r = methods[method](url, auth=auth, **kwargs)
@@ -59,18 +60,33 @@ def _request(auth, method, url, **kwargs):
     return response
 
 
+def _print_dict(_dict):
+    """
+    Prints a dict in a key=value format.
+
+    Args:
+        _dict (dict): values to be formatted
+    """
+    for k, v in _dict.items():
+        print '%s="%s"' % (k, v)
+
+
 class PullRequest:
     """
     https://developer.github.com/v3/pulls/
     """
-    def __init__(self, repo):
+    def __init__(self, repo, number=None):
         """
         Args:
             repo (str): git repository path. Format: <repo owner>/<repo>
+            number (int): pull-request number
         """
         self.repo = repo
         self._url = os.path.join(
             API_BASE_URL, "repos/%(repo)s/pulls" % dict(repo=self.repo))
+
+        if number:
+            self._set_attributes(self.get(number))
 
     def open(self, title, body, src_branch, dst_branch):
         """
@@ -93,6 +109,23 @@ class PullRequest:
         self._set_attributes(
             request('POST', self._url, data=json.dumps(payload_body)))
 
+    def get(self, pr_number):
+        """
+        Gets a GitHub pull-request
+
+        Args:
+            pr_number (int): pull-request number
+
+        Returns:
+            dict: pull-request information. See
+                https://developer.github.com/v3/pulls/#get-a-single-pull-request
+        """
+        url = os.path.join(self._url, str(pr_number))
+
+        pr_info = request('GET', url)
+
+        return pr_info
+
     def _set_attributes(self, pr_info):
         """
         Sets this class' attributes
@@ -100,6 +133,7 @@ class PullRequest:
         Args:
             pr_info (dict): information about a GitHub pull request
         """
+
         for k, v in pr_info.items():
             # just set the simple types that can later be accessed by
             # pull_request.attr. Ignore dicts for now because they
@@ -123,6 +157,24 @@ def open_pull_request(title, src_branch, repo, dst_branch, body=""):
     pr.open(title, body, src_branch, dst_branch)
 
 
+def query_pull_request(repo, number, **kwargs):
+    """
+    Handler for the query_pr command. Returns information about a
+    GitHub pull-request.
+
+    Args:
+        repo (str): target repository path. Format: <repo owner>/<repo>
+        number (int): pull-request number
+        kwargs: keys being queried in the form: key=True
+    """
+    pr = PullRequest(repo, number)
+
+    if all(not v for v in kwargs.values()):
+        kwargs['title'] = True
+
+    return {k:getattr(pr, k) for k, v in kwargs.items() if v}
+
+
 if __name__ == '__main__':
     commands = {
         'open_pr': {
@@ -137,6 +189,26 @@ if __name__ == '__main__':
                 (('dst_branch',), dict(metavar='dst-branch',
                     help='pull request destination branch: <branch>')),
                 (('--body',), dict(help='pull request body text')),
+            ],
+        },
+        'query_pr': {
+            'description': "Get information about a GitHub pull-request",
+            'function': query_pull_request,
+            'arguments': [
+                (('repo',), dict(
+                    help='target repository path: <repo owner>/<repo>')),
+                (('number',), dict(help='pull request number', type=int)),
+                (('--title',), dict(
+                    help='get pull request title text', action='store_true')),
+                (('--state',), dict(
+                    help='get pull request state, i.e. open, closed',
+                    action='store_true')),
+                (('--merged',), dict(
+                    help='get whether the pull request is merged',
+                    action='store_true')),
+                (('--mergeable',), dict(
+                    help='get whether the pull request can be merged',
+                    action='store_true')),
             ],
         },
     }
@@ -168,4 +240,6 @@ if __name__ == '__main__':
         if value is None:
             cmd_args.pop(key)
 
-    commands[args.command]['function'](**cmd_args)
+    cmd_output = commands[args.command]['function'](**cmd_args)
+    if cmd_output:
+        _print_dict(cmd_output)
--- a/scripts/github_api
+++ b/scripts/github_api
@@ -71,6 +71,40 @@ def _print_dict(_dict):
         print '%s="%s"' % (k, v)
 
 
+class Status:
+    """
+    https://developer.github.com/v3/repos/statuses/
+    """
+    def __init__(self, repo, reference):
+        """
+        Args:
+            repo (str): target repository path. Format:
+                <repo owner>/<repo>
+            reference (str): git reference or GitHub pull request
+                number
+        """
+        self.ref = reference
+        self._url = os.path.join(
+            API_BASE_URL,
+            "repos/%(repo)s/commits/%(reference)s/status" % dict(
+                repo=repo, reference=reference))
+        self.get()
+
+    def get(self):
+        """
+        Gets the status
+        """
+        status_info = request('GET', self._url)
+
+        self.num_contexts = status_info['total_count']
+
+        self.latest = {}
+        for status in status_info['statuses']:
+            context = status['context']
+            status.pop('context')
+            self.latest[context] = status
+
+
 class PullRequest:
     """
     https://developer.github.com/v3/pulls/
@@ -124,6 +158,8 @@ class PullRequest:
 
         pr_info = request('GET', url)
 
+        self.reference = pr_info['head']['sha']
+
         return pr_info
 
     def _set_attributes(self, pr_info):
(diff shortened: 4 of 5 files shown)


farosas:
This is on top of https://github.com/open-power-host-os/infrastructure/pull/29 please review only the last 4 commits: 

- Add query_pr command to GitHub API wrapper script (https://github.com/open-power-host-os/infrastructure/pull/31/commits/c5f8947404e8c314f632fe6df89b814935c56e20)
- Add query_status command to GitHub API wrapper script (https://github.com/open-power-host-os/infrastructure/pull/31/commits/91a3d5dfc5da39bf3b7408a1af092119429cc921)
- Add job to trigger ISO build from versions repo (https://github.com/open-power-host-os/infrastructure/pull/31/commits/fe9853a9ad03832b8f4fd5e580d56ae7fc01f47c)
- Alter trigger jobs to report triggered build URL in GitHub pull-request (https://github.com/open-power-host-os/infrastructure/pull/31/commits/3ea97b9d1a59976b61801dbe44e90ce96ecd90ee)

scripts/github_api
```diff
@@ -40,14 +41,22 @@ def _request(auth, method, endpoint, **kwargs):
     return response
 
 
+def _format_output(output_dict):
+    for k, v in output_dict.items():
+        print "%s=\"%s\"" % (k, v)
```
ghost: You can enclose the string in single quotes to avoid backslashes: `'%s="%s"'`

scripts/github_api
```diff
@@ -40,14 +41,22 @@ def _request(auth, method, endpoint, **kwargs):
...
+
+
 class PullRequest:
     """
     https://developer.github.com/v3/pulls/
     """
-    def __init__(self, repo):
+    def __init__(self, repo, number=None):
         self.repo = repo
         self._endpoint = API_BASE_URL + "repos/%(repo)s/pulls"
```
ghost: Why don't you replace `repo` here? It does not change.

scripts/github_api
```diff
@@ -86,6 +112,26 @@ if __name__ == '__main__':
...
+                    help='pull request title text', action='store_true')),
+                (('--state',), dict(
+                    help='pull request state, i.e. open, closed',
+                    action='store_true')),
+                (('--merged',), dict(
+                    help='whether the pull request is merged',
+                    action='store_true')),
+                (('--mergeable',), dict(
+                    help='whether the pull request can be merged',
+                    action='store_true')),
```
ghost: Mention those options are to choose what information is retrieved. E.g. add "retrieve" or "get" to the help text.

scripts/github_api
```diff
@@ -86,6 +112,26 @@ if __name__ == '__main__':
                 (('--body',), dict(help='pull request body text')),
             ],
         },
+        'query_pr': {
+            'description': "Get information about a GitHub pull-request",
+            'function': query_pull_request,
+            'arguments': [
+                (('repo',), dict(
+                    help='target repository path: <repo owner>/<repo>')),
+                (('number',), dict(help='pull request number')),
```
ghost: Use `type=int` to have some validation.

scripts/github_api
```diff
@@ -40,14 +41,22 @@ def _request(auth, method, endpoint, **kwargs):
     return response
 
 
+def _format_output(output_dict):
```
ghost: Either return a string and print outside or rename this to `_print_output`.

ghost (changes requested):

scripts/github_api
```diff
@@ -96,6 +123,28 @@ def query_pull_request(repo, number, **kwargs):
     return {k:getattr(pr, k) for k, v in kwargs.items() if v}
 
 
+def query_status(repo, reference, context, **kwargs):
+    try:
```
ghost: Comment that you are converting a PR number to a commit reference here.

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo.groovy
```diff
@@ -0,0 +1,57 @@
...
+      userWhitelist("${GHPRB_ADMIN_USER}")
+      orgWhitelist("${GHPRB_ADMIN_ORGANIZATION}")
+      allowMembersOfWhitelistedOrgsAsAdmin()
+      cron('H/5 * * * *')
+      triggerPhrase('.*start\\W+iso.*')
+      onlyTriggerPhrase()
+      extensions {
+	commitStatus {
+	  context('Build Host OS ISO')
+	  statusUrl('${JENKINS_URL}/job/build_host_os/${TRIGGERED_BUILD_NUMBER_build_host_os}')
```
ghost: `${JENKINS_URL}/job/build_host_os_iso/${TRIGGERED_BUILD_NUMBER_build_host_os_iso}`

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+set -e
+set -x
+
+
+TARGET_CONTEXT="Build Host OS"
```
ghost: "Build Host OS ISO"

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
```diff
@@ -0,0 +1,52 @@
...
+
+
+get_build_state(){
+    eval $(github_api "$GITHUB_USER_NAME" "$GITHUB_PASSWORD" \
+                      query_status "${GITHUB_ORGANIZATION_NAME}/versions" \
+                      "$ghprbPullId" "$TARGET_CONTEXT" --state \
+                      --target-url || echo "exit 1")
```
ghost: Isn't it better for the script to print something instead of placing this echo at every call?

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
```diff
@@ -23,6 +23,7 @@ job('trigger_host_os_build_from_builds_repo') {
       extensions {
 	commitStatus {
 	  context('Build Host OS')
+	  statusUrl('${JENKINS_URL}/job/build_host_os/${TRIGGERED_BUILD_NUMBER_build_host_os}')
```
ghost, line 26: This commit should come before the one that creates the trigger build ISO job.

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
farosas: This is the target context, the one that we're listening for change, so this is correct.

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
farosas: The echo exists only because of the 'eval'. There is nothing that the script could print (aside from "exit 1") that would make the whole script ~~subshell~~ exit. Moreover, this is only to force it to exit right after the failed command (github_api) and not in the next command that breaks (~~probably the eval itself,~~ since I'm using set -e)

EDIT: exit the whole script
EDIT2: eval won't break

ghost (approved):

scripts/github_api
```diff
@@ -59,19 +60,34 @@ def _request(auth, method, url, **kwargs):
     return response
 
 
+def _print_output(output):
```
alanoe: minor: I'd prefer _print_dict(_dict) and remove these "output" references. This function does not need to know it is printing output of the script

scripts/github_api
```diff
@@ -93,13 +109,30 @@ class PullRequest:
         self._set_attributes(
             request('POST', self._url, data=json.dumps(payload_body)))
 
+    def get(self, pr_number):
+        """
+        Gets a GitHub pull-request
```
alanoe, line 148: minor: I do not know why you chose to use this dash in "pull-request", I prefer to write the term as it is written in English ("pull request")

scripts/github_api
```diff
@@ -93,13 +109,30 @@ class PullRequest:
...
+
+        Args:
+            pr_number (str): pull-request number
```
alanoe: minor: if this is a number, it'd be better to pass an int (instead of str) to the method

scripts/github_api
```diff
@@ -93,13 +109,30 @@ class PullRequest:
...
+
+        Returns:
+            dict: pull-request information
```
alanoe: What are the valid keys of this dict? Either point to that info in GitHub API doc or mention it explicitly.

scripts/github_api
```diff
@@ -93,13 +109,30 @@ class PullRequest:
...
+        """
+        endpoint = os.path.join(self._url, pr_number)
```
alanoe: I'd prefer to not use API-specific terminology (eg HTTP requests have a URL, not an endpoint)

scripts/github_api
```diff
@@ -123,6 +156,24 @@ def open_pull_request(title, src_branch, repo, dst_branch, body=""):
...
+    """
+    Handler for the query_pr command. Returns information about a
+    GitHub pull-request.
+
+    Args:
+        repo (str): target repository path. Format: <repo owner>/<repo>
+        number (int): pull-request number
+        kwargs: keys being queried in the form: key=True
+    """
+    pr = PullRequest(repo, str(number))
```
alanoe: If you make PullRequest class constructor accept an int as PR number, you do not need to cast it here

scripts/github_api
```diff
@@ -123,6 +156,24 @@ def open_pull_request(title, src_branch, repo, dst_branch, body=""):
...
 
+def query_pull_request(repo, number, **kwargs):
+    """
+    Handler for the query_pr command. Returns information about a
+    GitHub pull-request.
+
+    Args:
+        repo (str): target repository path. Format: <repo owner>/<repo>
+        number (int): pull-request number
+        kwargs: keys being queried in the form: key=True
```
alanoe, line 204: Why not simply use a list with the queried fields names (as in the SELECT clause of a relational db query)? This would be much easier to the user of this Python method IMHO. 

Maybe you are using this format to make script's command line arguments easier to parse? I'd prefer to not couple GitHub API wrapper and the command line parsing, this will make this code harder to use out of the command line in the future.

scripts/github_api
```diff
@@ -139,6 +190,26 @@ if __name__ == '__main__':
                 (('--body',), dict(help='pull request body text')),
             ],
         },
+        'query_pr': {
+            'description': "Get information about a GitHub pull-request",
+            'function': query_pull_request,
+            'arguments': [
+                (('repo',), dict(
+                    help='target repository path: <repo owner>/<repo>')),
```
alanoe, line 271: This is a little confusing. I assume the text after the colon (:) is the path format. Pls mention it explicitly. eg "target repository path. Format: `<repo owner`>/`<repo`>"

scripts/github_api
```diff
@@ -71,6 +71,40 @@ def _print_output(output):
...
+    """
+    https://developer.github.com/v3/repos/statuses/
+    """
+    def __init__(self, repo, reference):
+        """
+        Args:
+            repo (str): target repository path. Format:
+                <repo owner>/<repo>
+            reference (str): git reference. Either a sha1 or a
+                pull-request number
```
alanoe: - I'd prefer to not use a dash ("-") in "pull-request"
- A pull request number is not a git reference. Maybe you meant "a git reference (sha1 commit ID, tag name or branch name) or a GitHub pull request number"
- Do we need to accept a git reference? We'll always want to query and update a pull request status and we'll have its number

scripts/github_api
```diff
@@ -71,6 +71,40 @@ def _print_output(output):
...
+        """
+        self.ref = reference
+        self._endpoint = os.path.join(
```
alanoe: I'd prefer to not use API-specific terminology, but I have already mentioned this other times. I'll leave it up to you.

scripts/github_api
```diff
@@ -71,6 +71,40 @@ def _print_output(output):
         print '%s="%s"' % (k, v)
 
 
+class Status:
+    """
+    https://developer.github.com/v3/repos/statuses/
+    """
+    def __init__(self, repo, reference):
```
alanoe, line 78: It makes more sense to me to have only a pull request param (from which you may infer repo and reference)

scripts/github_api
```diff
@@ -174,6 +210,42 @@ def query_pull_request(repo, number, **kwargs):
...
+        # if we were given a number as the reference, interpret is as
+        # a pull-request number
+        int(reference)
+        pr = PullRequest(repo, reference)
+        reference = pr.reference
+    except:
+        pass
+
+    status = Status(repo, reference)
+    if not status.num_contexts or context not in status.latest:
```
alanoe, line 237: imho it is clearer to compare status.num_contexts to 0

scripts/github_api
```diff
@@ -123,6 +157,8 @@ class PullRequest:
 
         pr_info = request('GET', endpoint)
 
+        self.reference = os.path.basename(pr_info['statuses_url'])
```
alanoe: - Not clear to me what you named this "reference" instead of "statuses_url"
- I was expecting to see a variable in the PullRequest object which keeps its statuses. Why there isn't one?

scripts/github_api
```diff
@@ -59,19 +60,34 @@ def _request(auth, method, url, **kwargs):
...
+    Prints output from this script in a key=value format.
+
+    Args:
+        output (dict): values to be formatted
+    """
+    for k, v in output.items():
+        print '%s="%s"' % (k, v)
+
+
 class PullRequest:
```
alanoe: not related exactly to this patch: It seems to me this class is not a PullRequest, it is a PullRequestManager, as it creates (open() method) PRs and queries them (get() method). I'd prefer to name this better to avoid confusion.

scripts/github_api
```diff
@@ -174,6 +210,42 @@ def query_pull_request(repo, number, **kwargs):
...
+    """
+    Handler for the query_status command. Returns information about
+    the status of a GitHub issue or pull-request.
+
+    Args:
+        repo (str): target repository path. Format:
+            <repo owner>/<repo>
+        reference (str): git reference. Either a sha1 or a
+            pull-request number
+        context (str): status context to query
```
alanoe, line 224: not sure how'll we use this "context". Are Pylint validation and OP Host OS package building separate contexts in a builds git repo PR?

scripts/github_api
```diff
@@ -174,6 +210,42 @@ def query_pull_request(repo, number, **kwargs):
     return {k:getattr(pr, k) for k, v in kwargs.items() if v}
 
 
+def query_status(repo, reference, context, **kwargs):
```
alanoe: I dislike using this kwargs instead of having a simple list

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
```diff
@@ -23,6 +23,7 @@ job('trigger_host_os_build_from_builds_repo') {
       extensions {
```
alanoe, line 23: - From commit msg, it seems you'll update all trigger jobs, but you only touch the ones that build packages. Why?
- "include the trigger build URL in their "vote"."  . Not sure how this affects the PR. GitHub creates a link with this URL in the PR web page? If so, mention in commit longer msg to make things clearer.

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
alanoe, line 26: I have no idea where TRIGGERED_BUILD_NUMBER_build_host_os comes from, you could mention it somewhere

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
alanoe: I suggest to use clearer terminology, such as "LISTENED_CONTEXT"

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
```diff
@@ -0,0 +1,56 @@
...
+get_pr_state(){
+    # this will evaluate the queried keys as variables: state=<state>,
+    # title=<title>
+    eval $(github_api "$GITHUB_USER_NAME" "$GITHUB_PASSWORD" \
+                      query_pr "${GITHUB_ORGANIZATION_NAME}/versions" \
+                      "$ghprbPullId" --state --title || echo "exit 1")
+}
+
+state="pending"
+get_build_state
```
alanoe, line 28: it is strange that a get_* function does not return anything. It sets the global variable named "state" ? @@

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
```diff
@@ -0,0 +1,56 @@
...
+        echo "Pull request ${ghprbPullId}: \"${title}\" was closed, aborting."
+        echo "$ghprbPullLink"
+        exit 1
+    fi
+fi
+
+echo "BUILD_JOB_NUMBER=$(basename $target_url)" > BUILD_PARAMETERS
+echo "Triggering ISO build from $target_url ..."
+
+exit 0
```
alanoe, line 56: side-note: it seems to me we are writing all the code which integrates our Jenkins jobs and GitHub API in shell script, IMHO we should consider doing it in Python, it would make the code much simpler and make maintenance easier

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
```diff
@@ -0,0 +1,56 @@
...
+TARGET_CONTEXT="Build Host OS"
+
+
+get_build_state(){
+    # this will evaluate the queried keys as variables: state=<state>,
+    # target_url=<url>
+    eval $(github_api "$GITHUB_USER_NAME" "$GITHUB_PASSWORD" \
+                      query_status "${GITHUB_ORGANIZATION_NAME}/versions" \
+                      "$ghprbPullId" "$TARGET_CONTEXT" --state \
+                      --target-url || echo "exit 1")
```
alanoe: I'd avoid using positional arguments. Also, IMHO using a fields/attributes/keys argument with desired keys in comma-separated format would be simpler. Eg:

github_api --user $GITHUB_USER_NAME --password $ GITHUB_PASSWORD
      query_status --repo "${GITHUB_ORGANIZATION_NAME}/versions" 
      --pull-request "$ghprbPullId" --context "$TARGET_CONTEXT" --fields "state,target-url"

alanoe (changes requested):

scripts/github_api
farosas: ok, I'll keep as it is for now until I finish the script though

scripts/github_api
farosas: Just trying to make it obvious that this is a specific thing and not any "request to pull". There's code already merged with the dash so there's not much I can do now.

scripts/github_api
farosas: First, this method is in the "command line part" of the script. I don't expect it to be called from python scripts. I'll send a patch after I finish the code changing the names of these handlers to something more obvious.

That said, I think having the options explicitly listed in the command line using argparse more maintainable and usable than a huge text with "List of values to query. Possible values: a,b,c,d." and the doc for each possible value. I also like the fact that adding a new possible value to be queried touches only the argparse code.

scripts/github_api
farosas: I don't get it. The text after the colon is indeed the path format.

scripts/github_api
farosas: The status API is linked to a reference, not a pull request. I don't want to make it pull request specific.

scripts/github_api
farosas: * this will be done later
* yes, that's what I meant
* I don't see the need to couple this to pull requests. Accepting a reference makes it more true to the API and allow this class to be used in other situations too.

scripts/github_api
farosas: * That is the reference to the PRs head commit, which afaik is how GitHub references the pull request. I just found a clearer way to access it: pr_info['head']['sha']
* The pull requests API endpoint only provides the link to the statuses API endpoint for that PR. I haven't created a variable in PullRequest with the status because 1) since the status API is not used only for pull requests I can have an uniform way of accessing it for any type of reference and 2) getting the status would force another API call to be made at every instantiation of PullRequest. By accessing the Status object directly we avoid the extra call when the user only wants PR information and not PR Status information.

scripts/github_api
farosas: Yes. We currently define this in the Jenkins GitHub pull request builder plugin. This is used to identify which status is being queried, e.g. to get the status of a Pylint validation

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
farosas: * they are the only ones that use the GitHub pull request build plugin and therefore the only ones that have any context to set
* yes

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
farosas: I was about to do this but then I realized all the variables from this file come from Jenkins so I don't see why give this one special treatment.

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
farosas: I'll continue to work on the infrastructure code after this feature is in. I'll investigate and possibly port our jobs to use the Jenkins Pipeline [1]. That will free us from this entangling of jobs and (sometimes broken) plugins we've been making. Along with that I intend to drop shell scripts where feasible. Since I'm almost done with these patches I'd like to merge them and work on that later.


1- https://jenkins.io/doc/book/pipeline/getting-started/

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
farosas: Well, there won't be a word which suits this situation. Yes, it will create a global variable like it's documented in line 20

scripts/github_api
alanoe: AV> Just trying to make it obvious that this is a specific thing and not any "request to pull". There's code already merged with the dash so there's not much I can do now.

Sorry, but the dash don't make that obvious to me *at all*, I think readability is better without it. Of course you can do, you just need to checkout master, create a branch, replace all occurrences of 'pull-request' with 'pull request', submit a patch and rebase this PR on top of it. It takes a few minutes to do it. Up to you, I wont fight over it.

scripts/github_api
alanoe: FR> yes, that's what I meant

Pls fix doc then

PR> I don't see the need to couple this to pull requests. Accepting a reference makes it more true to the API and allow this class to be used in other situations too.

1) I personally prefer a clearer OO model than trying to mimick 1-1 the underlying API. We have talked about this when you were working in the Blivet wrapper in PowerKVM.  

2) IMHO the data representation would be clearer and easier to understand in the API wrapper if there was a 1-1 reference between Pull Request and PullRequestStatus. A GitHub pull request status does not make sense without a pull request IMHO, what this GitHub API call with the git reference probably does is find the PR number from the git reference. 

3) we do not have any need to create status from a git reference AFAIK.  

4) *If* we need such feature in the future, we could simply create an util function (not inside any class) to do that.

scripts/github_api
alanoe: Got it. Thanks for explaining it.

I'd prefer GitHubCheckStatus instead of status and status_name instead of context, those names uses GitHub web UI terminology and are much more clearer to understand. Up to you

scripts/github_api
alanoe: We disagree on the design, but I see your point. Never mind then.

scripts/github_api
alanoe: Unsure of what happened here, I probably reviewed an older version. Never mind.

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
alanoe: AE> I'd avoid using positional arguments. Also, IMHO using a fields/attributes/keys argument with desired keys in comma-separated format would be simpler

As you have not commented on this, I assume you disagree. Too bad

jenkins_jobs/trigger_host_os_iso_build_from_versions_repo/script.sh
alanoe: I just remembered you cannot return a string in shell script. I really dislike shell script!

http://stackoverflow.com/questions/3236871/how-to-return-a-string-value-from-a-bash-function
2nd answer suggests passing a variable "by-reference" and altering it inside the function (similar to C). It seems clearer to me, at least you are altering a function parameter and not *any* global script variable you choose.

jenkins_jobs/trigger_host_os_build_from_builds_repo.groovy
alanoe: OK. (1) Not knowing if the variable is a Jenkins job variable, a global Jenkins custom variable or a global Jenkins implicit variable and (2) having no doc about the meaning of these variables in the script (in this case, groovy script) which uses them really annoy me, but never mind.

PR #32, opened 2017-02-15
Add a Jenkins job which builds old versions

alanoe:


farosas (approved):

farosas:
There is no setup of any trigger for this job. When will it run?

EDIT: I'll merge it and we can think about the trigger later

PR #34, opened 2017-02-17
Turn uploads into build steps instead of post build steps

ghost:
This requires archiving the build results with a custom script, because
the Jenkins archive plugin can only be executed as a post build step.
Artifacts that are not uploaded (e.g. logs) can still be kept as a post
build step.

This allows us to wait for the upload to finish and mark the build as
unstable if the upload fails.
It avoids creating symbolic links prematurely in GSA for weekly and
nightly builds, which could allow users to access incomplete
repositories.

### DIFF
--- a/jenkins_jobs/build_host_os/archive.sh
+++ b/jenkins_jobs/build_host_os/archive.sh
@@ -1,0 +1,18 @@
+JENKINS_MASTER=$(basename $JENKINS_URL)
+JENKINS_BUILD_ARCHIVE_DIR_NAME="archive"
+JENKINS_BUILD_ARCHIVE_URL=${JENKINS_MASTER}:${JENKINS_HOME}/jobs/${JOB_NAME}/builds/${BUILD_NUMBER}/${JENKINS_BUILD_ARCHIVE_DIR_NAME}
+
+rsync_upload() {
+    rsync -e 'ssh -i ${HOME}/.ssh/jenkins_id_rsa' \
+              --verbose --compress --stats --times --perms \
+              $@ $JENKINS_BUILD_ARCHIVE_URL
+}
+
+
+# Create remote archive directory
+mkdir $JENKINS_BUILD_ARCHIVE_DIR_NAME
+rsync_upload --recursive $JENKINS_BUILD_ARCHIVE_DIR_NAME
+
+rsync_upload SUCCESS
+rsync_upload BUILD_TIMESTAMP
+rsync_upload --recursive repository
--- a/jenkins_jobs/trigger_nightly_host_os_build.groovy
+++ b/jenkins_jobs/trigger_nightly_host_os_build.groovy
@@ -42,9 +42,8 @@ job('trigger_nightly_host_os_build') {
     downstreamParameterized {
       trigger('build_host_os') {
 	block {
-	  buildStepFailure('FAILURE')
-	  failure('FAILURE')
-	  unstable('UNSTABLE')
+          buildStepFailure('UNSTABLE')
+          failure('UNSTABLE')
 	}
 	parameters {
 	  predefinedProps(
@@ -69,9 +68,8 @@ job('trigger_nightly_host_os_build') {
     downstreamParameterized {
       trigger('build_host_os_iso') {
         block {
-          buildStepFailure('FAILURE')
-          failure('FAILURE')
-          unstable('UNSTABLE')
+          buildStepFailure('UNSTABLE')
+          failure('UNSTABLE')
         }
         parameters {
           predefinedProps(
--- a/jenkins_jobs/upload_build_artifacts/script.sh
+++ b/jenkins_jobs/upload_build_artifacts/script.sh
@@ -17,11 +17,11 @@ alias rsync_upload="rsync -e \
     --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"
 
 # Create remote build directory
-mkdir ${BUILD_TIMESTAMP}
-rsync_upload --recursive ./${BUILD_TIMESTAMP} ${BUILD_DIR_RSYNC_URL}
+mkdir $BUILD_TIMESTAMP
+rsync_upload --recursive ./$BUILD_TIMESTAMP $BUILD_DIR_RSYNC_URL
 
-rsync_upload STATUS ${BUILD_DIR_RSYNC_URL}
-rsync_upload --recursive repository ${BUILD_DIR_RSYNC_URL}
+rsync_upload STATUS $BUILD_DIR_RSYNC_URL
+rsync_upload --recursive repository $BUILD_DIR_RSYNC_URL
 
 # Create hostos.repo
 echo -e """[hostos]
@@ -31,4 +31,4 @@ enabled=1
 priority=1
 gpgcheck=0""" > hostos.repo
 
-rsync_upload hostos.repo ${BUILD_DIR_RSYNC_URL}
+rsync_upload hostos.repo $BUILD_DIR_RSYNC_URL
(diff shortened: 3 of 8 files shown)


jenkins_jobs/build_host_os/archive.sh
```diff
@@ -0,0 +1,16 @@
+JENKINS_MASTER=$(basename $JENKINS_URL)
+JENKINS_BUILD_ARCHIVE_DIR_NAME="archive"
+JENKINS_BUILD_ARCHIVE_URL=${JENKINS_MASTER}:${JENKINS_HOME}/jobs/${JOB_NAME}/builds/${BUILD_NUMBER}/${JENKINS_BUILD_ARCHIVE_DIR_NAME}
+
+alias rsync_upload="rsync -e \
+    'ssh -i ${HOME}/.ssh/jenkins_id_rsa' \
+    --verbose --compress --stats --times --perms"
```
farosas: I think using a function instead of an alias would improve legibility

```
rsync_upload() {
    rsync -e 'ssh -i ${HOME}/.ssh/jenkins_id_rsa' \
              --verbose --compress --stats --times --perms \
              $@ $JENKINS_BUILD_ARCHIVE_URL
}

rsync_upload --recursive $JENKINS_BUILD_ARCHIVE_DIR_NAME
rsync_upload SUCCESS
rsync_upload BUILD_TIMESTAMP
rsync_upload --recursive repository
```

ghost: That's better. Done it for upload jobs too.

alanoe:
OH> This requires archiving the build results manually

This is ambiguous. "manually" usually means that "someone must do it interactively", but I think you mean that "our code must automatically do it, without using Jenkins archive plugin". Could you make it clearer?

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -12,16 +12,18 @@ else
...
-    'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
-    --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"
+rsync_upload() {
+    rsync -e 'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
+              --verbose --compress --stats --times --chmod=a+rwx,g+wx,o- \
+              $@ $BUILD_DIR_RSYNC_URL
+}
 
 # Create remote build directory
-mkdir ${BUILD_TIMESTAMP}
```
alanoe, line 20: I'd remove curly braces in a separate patch and explain why

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -12,16 +12,18 @@ else
 fi
 echo  "{'REPO': '$REPOSITORY_FILE_URL', 'BUILD_TIMESTAMP': '${BUILD_TIMESTAMP}', 'BUILD_LOG': '$BUILD_URL', 'BUILD_ID': $BUILD_JOB_NUMBER, 'BUILD_STATUS': '$BUILD_STATUS'}" > STATUS
 
-alias rsync_upload="rsync -e \
-    'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
-    --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"
+rsync_upload() {
```
alanoe, line 15: Could you document what this function expects as parameter (source file path) and mention that destination is hardcoded?

jenkins_jobs/upload_build_artifacts/script.sh
```diff
@@ -12,16 +12,18 @@ else
 fi
 echo  "{'REPO': '$REPOSITORY_FILE_URL', 'BUILD_TIMESTAMP': '${BUILD_TIMESTAMP}', 'BUILD_LOG': '$BUILD_URL', 'BUILD_ID': $BUILD_JOB_NUMBER, 'BUILD_STATUS': '$BUILD_STATUS'}" > STATUS
 
-alias rsync_upload="rsync -e \
-    'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
```
alanoe, line 16: not related to this patch: I have no idea where HOME and UPLOAD_SERVER_USER_NAME come from, Jenkins environment variables? Could we mention those somewhere in the script?

jenkins_jobs/upload_iso/script.sh
```diff
@@ -3,13 +3,15 @@ BUILD_DIR_PATH="${UPLOAD_SERVER_BUILDS_DIR}/${BUILD_TIMESTAMP}"
 BUILD_DIR_RSYNC_URL="${UPLOAD_SERVER_USER_NAME}@${UPLOAD_SERVER_HOST_NAME}:${BUILD_DIR_PATH}"
 ISO_DIR_RSYNC_URL="${BUILD_DIR_RSYNC_URL}/iso"
 
-alias rsync_upload="rsync -e \
-    'ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa' \
-    --verbose --compress --stats --times --chmod=a+rwx,g+wx,o-"
+rsync_upload() {
```
alanoe, line 6: Same comments of the other script apply here

jenkins_jobs/trigger_nightly_host_os_build.groovy
```diff
@@ -69,9 +68,8 @@ job('trigger_nightly_host_os_build') {
     downstreamParameterized {
       trigger('build_host_os_iso') {
         block {
-          buildStepFailure('FAILURE')
-          failure('FAILURE')
-          unstable('UNSTABLE')
+          buildStepFailure('UNSTABLE')
+          failure('UNSTABLE')
```
alanoe, line 72: It seems to me this sets the build as "unstable" even if build fails. If so, this does not make sense. Wouldn't it be better to split ISO build and upload in separate jobs and react differently to a failure in each of them?

alanoe (changes requested):

jenkins_jobs/trigger_nightly_host_os_build.groovy
ghost: Those names indeed do not say much. From the Jenkins interface, you can see:
`buildStepFailure('UNSTABLE')` means "Fail this build step if the triggered build is worse or equal to UNSTABLE"
and `failure('UNSTABLE')` means "Mark this build as failure if the triggered build is worse or equal to UNSTABLE"

jenkins_jobs/upload_build_artifacts/script.sh
ghost: All variables come from the environment. Some are unique to the Jenkins instance (HOME), some come from job parameters (UPLOAD_SERVER_USER_NAME) and some come from plugins. I don't think documenting every variable is worth the trouble.

jenkins_jobs/upload_build_artifacts/script.sh
ghost: Ok.

jenkins_jobs/upload_build_artifacts/script.sh
ghost: I think this can be easily deduced from looking at the single command it executes.
Do we really need to add this extra documentation? This was supposed to simplify the script, I can revert back if you think it doesn't.

jenkins_jobs/upload_build_artifacts/script.sh
alanoe: Having to read entire script to assert if the variable is defined somewhere or if it comes from environment variables is cumbersome and makes maintenance harder IMHO. If I am the only one who thinks so, never mind.

jenkins_jobs/upload_build_artifacts/script.sh
alanoe: The point of using a function is that you should not need to read its body (implementation) to understand what it does, only its header. I prefer the function than the alias, but I see no problem in adding one or two lines of doc to make code reading easier. We document all functions in Python, why not do the same in shell script? If you think it does not add value, never mind.

farosas (approved):

PR #46, opened 2017-03-01
Allow the use of custom 'versions' url/branch during periodic builds

farosas:
Currently the upgrade_versions command does not receive the 'versions'
repository url and branch via parameter. This is not ideal because in
a development environment using a custom GitHub user/organization, the
command would still clone open-power-host-os/versions and that is
misleading.


ghost (changes requested):
The same should be done for nightly builds

ghost (approved):

PR #55, opened 2017-03-29
Fix shell script of Jenkins job which build from versions git repository old commits

alanoe:
- Fix build-package subcommand name in host_os.py. The subcommand name
  was updated in builds git repository, but it was not updated here.
- Fix attributes names in the string replacement in config.yaml


ghost (approved):

mopsfelder:
Good catch, @alanoe.  Thank you!

farosas (approved):

PR #63, opened 2017-04-12
Fix versions repository URL passed to the create-release-notes command

ghost:
This was broken since cea30f93947610b5a817ccea76aebed661fe0647, but did
not show up because the versions repository was shared among
"update-versions" and "create-release-notes" and so the the command
"git checkout weekly-<timestamp>" had no problems in finding the remote
reference.

### DIFF
--- a/jenkins_jobs/trigger_weekly_host_os_build/script.sh
+++ b/jenkins_jobs/trigger_weekly_host_os_build/script.sh
@@ -77,7 +77,7 @@ create_release_notes() {
     python host_os.py \
            --verbose \
            build-release-notes \
-               --packages-metadata-repo-url "$VERSIONS_REPOSITORY_URL" \
+               --packages-metadata-repo-url "$VERSIONS_PUSH_URL" \
                --packages-metadata-repo-branch "$COMMIT_BRANCH" \
                --updater-name "$GITHUB_BOT_NAME" \
                --updater-email "$GITHUB_BOT_EMAIL" \


jenkins_jobs/trigger_weekly_host_os_build/script.sh
```diff
@@ -77,7 +77,7 @@ create_release_notes() {
     python host_os.py \
            --verbose \
            build-release-notes \
-               --packages-metadata-repo-url "$VERSIONS_REPOSITORY_URL" \
+               --packages-metadata-repo-url "$VERSIONS_PUSH_URL" \
```
farosas, line 80: This doesn't seem right to me. The push URL is the forked repo of the user doing the update (the bot in the general case), I don't think we should be reading from this repo, it might not even be up-to-date.

update-versions uses the upstream repo for fetching https://github.com/olavphibm/infrastructure/blob/a73ee0d6ec53074af3ea7d0d4fc6380532b9d25d/jenkins_jobs/trigger_weekly_host_os_build/script.sh#L68

farosas (changes requested):

jenkins_jobs/trigger_weekly_host_os_build/script.sh
ghost: That's right, `update-versions` fetches from the upstream repo, but it creates a new commit and pushes it to `VERSIONS_PUSH_URL`. This commit is then used to create the pull request. It will be up-to-date because it was based on the upstream's master, not on the fork's master.

farosas: I see. Thanks for clarifying.

farosas (approved):

ghost:
This fix has a small issue, I'm working on a better way to fix it.

ghost:
This has been treated in https://github.com/open-power-host-os/infrastructure/pull/65/commits/0454436408a2ae2fc44bbead95217feb23d7dad7

PR #65, opened 2017-04-17
Do not clone versions repository

ghost:
Depends on https://github.com/open-power-host-os/builds/pull/245

The builds script is now configured by default to fetch GitHub pull
requests for the versions repository, so we do not need this special
handling here.

Some fixes and refactorings had to be done to allow this.

### DIFF
--- a/jenkins_jobs/build_host_os/script.sh
+++ b/jenkins_jobs/build_host_os/script.sh
@@ -10,22 +10,6 @@ MAIN_CENTOS_REPO_RELEASE_URL="http://mirror.centos.org/altarch/7"
 # Jenkins handle its cleanup, since it's in the job workspace.
 mkdir mock_build
 
-# Set origin remote URL
-# This is the remote name assumed by the GHPRB plugin
-if [ -d $VERSIONS_REPO_PATH/.git ]; then
-    pushd $VERSIONS_REPO_PATH
-    git remote remove origin || true
-    git remote add origin $VERSIONS_REPO_URL
-else
-    git clone $VERSIONS_REPO_URL $VERSIONS_REPO_PATH --no-checkout
-    pushd $VERSIONS_REPO_PATH
-fi
-
-# Fetch pull requests in case this job was triggered by one
-git fetch origin +refs/pull/*:refs/remotes/origin/pr/*
-
-popd
-
 # Tell mock to use a different mirror/repo. This could be used to:
 # - speedup the chroot installation
 # - use a different version of CentOS
--- a/jenkins_jobs/build_host_os/script.sh
+++ b/jenkins_jobs/build_host_os/script.sh
@@ -10,11 +10,6 @@ MAIN_CENTOS_REPO_RELEASE_URL="http://mirror.centos.org/altarch/7"
 # Jenkins handle its cleanup, since it's in the job workspace.
 mkdir mock_build
 
-# Resolve the reference name to a commit id. If the value is not a
-# reference, assume it is a commit id.
-builds_repo_commit=$(git show-ref --hash "$BUILDS_REPO_REFERENCE" \
-                         || echo "$BUILDS_REPO_REFERENCE")
-
 # Set origin remote URL
 # This is the remote name assumed by the GHPRB plugin
 if [ -d $VERSIONS_REPO_PATH/.git ]; then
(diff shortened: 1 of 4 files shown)


jenkins_jobs/trigger_weekly_host_os_build/script.sh
```diff
@@ -1,16 +1,17 @@
 set -e
 
-VERSIONS_REPOSITORY_URL="https://github.com/${GITHUB_ORGANIZATION_NAME}/versions.git"
-PUSH_URL_PREFIX="ssh://git@github.com/${GITHUB_BOT_USER_NAME}"
+MAIN_REPO_URL_PREFIX="https://github.com/${GITHUB_ORGANIZATION_NAME}"
```
farosas: This is ambiguous. (main repo) url or main (repo url) ? Isn't a "repo url" a remote?

jenkins_jobs/trigger_weekly_host_os_build/script.sh
```diff
@@ -1,17 +1,23 @@
 set -e
 
-MAIN_REPO_URL_PREFIX="https://github.com/${GITHUB_ORGANIZATION_NAME}"
+MAIN_REPO_URL_PREFIX="ssh://git@github.com/${GITHUB_ORGANIZATION_NAME}"
 PUSH_REPO_URL_PREFIX="ssh://git@github.com/${GITHUB_BOT_USER_NAME}"
 
+BUILDS_WORKSPACE_DIR="/var/lib/host-os"
```
farosas: See line 22

jenkins_jobs/build_host_os/script.sh
```diff
@@ -29,8 +24,6 @@ fi
 # Fetch pull requests in case this job was triggered by one
 git fetch origin +refs/pull/*:refs/remotes/origin/pr/*
```
farosas: Shouldn't this be removed?

jenkins_jobs/trigger_weekly_host_os_build/script.sh
```diff
@@ -1,17 +1,23 @@
...
+REPOSITORIES_PATH="${BUILDS_WORKSPACE_DIR}/repositories"
+
 VERSIONS_REPO_NAME="versions"
 VERSIONS_MAIN_REPO_URL="${MAIN_REPO_URL_PREFIX}/${VERSIONS_REPO_NAME}.git"
 VERSIONS_PUSH_REPO_URL="${PUSH_REPO_URL_PREFIX}/${VERSIONS_REPO_NAME}.git"
+VERSIONS_REPO_PATH="${REPOSITORIES_PATH}/${VERSIONS_REPO_NAME}"
```
farosas: Why isn't this `versions_update-versions`?

farosas (changes requested):

jenkins_jobs/trigger_weekly_host_os_build/script.sh
ghost: A remote refers to a repository and may actually contain multiple URLs.
Using `MAIN_REMOTE_PREFIX` is kinda ambiguous too, since we are not specifying which part of the remote the prefix is refering to (name, URL, branch).
We could use `URL_PREFIX_TO_MAIN_REPO` maybe? Most variables would have to be renamed though. I don't have any better suggestions.

jenkins_jobs/trigger_weekly_host_os_build/script.sh
ghost: oops

jenkins_jobs/trigger_weekly_host_os_build/script.sh
ghost: This is done separately in the next commit.

jenkins_jobs/build_host_os/script.sh
ghost: It is removed in the next commit: "Do not clone versions repository"
I thought separating those commits would make it easier to see what exactly was removed and what just changed place.

jenkins_jobs/trigger_weekly_host_os_build/script.sh
farosas: Meh, where's the bicycle emoji when we need it?

jenkins_jobs/build_host_os/script.sh
farosas: No, that's not it; I think GitHub messed with the commit order again. Nevermind then.

farosas (approved):

jenkins_jobs/trigger_weekly_host_os_build/script.sh
ghost: :bike:shedding

PR #67, opened 2017-04-19
Update build directory path

mopsfelder:
Rename build directory path to `builds`, a more meaningful name to describe the
place where all builds are uploaded to.

Signed-off-by: Murilo Opsfelder Araujo <muriloo@linux.vnet.ibm.com>


ghost (approved):
We'll have to rename the dir in GSA, set the seed job default path and re-run it.
Ideally, this path should not be hardcoded in the scripts like this.

PR #75, opened 2017-05-18
Specify iso version when building iso

mopsfelder:
Depends on https://github.com/open-power-host-os/builds/pull/262

This specifies iso version to the build_host_os_iso job; and updates rsync
params of the upload_iso job not to replace existing files on receiver.

### DIFF
--- a/jenkins_jobs/build_host_os_iso/script.sh
+++ b/jenkins_jobs/build_host_os_iso/script.sh
@@ -1,8 +1,16 @@
 VERSIONS_REPO_DIR=$(basename $VERSIONS_REPO_URL .git)
 BUILDS_CONFIG_FILE="config/host_os.yaml"
+BUILD_TIMESTAMP=$(cat BUILD_TIMESTAMP)
+ISO_VERSION=$BUILD_TIMESTAMP
 MOCK_CONFIG_FILE="config/mock/CentOS/7/build-iso-CentOS-7-ppc64le.cfg"
 MAIN_CENTOS_REPO_RELEASE_URL="http://mirror.centos.org/altarch/7"
 MAIN_EPEL_REPO_RELEASE_URL="http://download.fedoraproject.org/pub/epel/7"
+
+# Format iso version.  Example of expected string:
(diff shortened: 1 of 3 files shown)


jenkins_jobs/upload_iso/script.sh
```diff
@@ -6,6 +6,7 @@ ISO_DIR_RSYNC_URL="${BUILD_DIR_RSYNC_URL}/iso"
 rsync_upload() {
     rsync -e "ssh -i ${HOME}/.ssh/${UPLOAD_SERVER_USER_NAME}_id_rsa" \
               --verbose --compress --stats --times --chmod=a+rwx,g+wx,o- \
+              --ignore-existing --itemize-changes \
```
olavph, line 9: I'm sorry for suggesting the use of this parameter, but since we already have `--verbose`, `--itemize-changes` does not add any useful information. The files that are sent are already listed, and the extra update information given by `--itemize-changes` will never exist, since we are ignoring existent files.

jenkins_jobs/build_host_os_iso/script.sh
```diff
@@ -1,9 +1,16 @@
 VERSIONS_REPO_DIR=$(basename $VERSIONS_REPO_URL .git)
 BUILDS_CONFIG_FILE="config/host_os.yaml"
+BUILD_TIMESTAMP=$(cat BUILD_TIMESTAMP)
+ISO_VERSION=$BUILD_TIMESTAMP
 MOCK_CONFIG_FILE="config/mock/CentOS/7/build-iso-CentOS-7-ppc64le.cfg"
 MAIN_CENTOS_REPO_RELEASE_URL="http://mirror.centos.org/altarch/7"
 MAIN_EPEL_REPO_RELEASE_URL="http://download.fedoraproject.org/pub/epel/7"
 
+# format iso version
```
farosas: It would be nice to add some doc about what is the expected input string.

farosas (approved):

olavph (approved):

PR #77, opened 2017-05-29
Create Jenkins pipelines

olavph:


jenkins_jobs/build_pipelines.groovy
```diff
@@ -0,0 +1,29 @@
...
+        buildForkPRHead(true)
+        buildForkPRMerge(false)
+        buildOriginBranch(false)
+        buildOriginBranchWithPR(false)
+        buildOriginPRHead(true)
+        buildOriginPRMerge(false)
+      }
+    }
+    triggers {
+      periodic(1)
```
farosas, line 17: This has a periodic trigger but `createPeriodicPipeline` don't, is that correct?

jenkins_jobs/periodic_pipelines.groovy
```diff
@@ -0,0 +1,23 @@
...
+  dir('infrastructure') {
+    git(url: 'ssh://git@github.com/$GITHUB_ORGANIZATION_NAME/infrastructure.git',
+        branch: 'pipeline')
+  }
+  pipeline = load 'infrastructure/pipeline/${name}.groovy'
+}
+
+pipeline.execute()
+""")
+        sandbox()
```
farosas, line 16: What does this do?

jenkins_jobs/build_pipelines.groovy
olavph: Yes. This trigger is for the polling of changes in the repositories.

jenkins_jobs/periodic_pipelines.groovy
olavph: It allows the execution of the script, as long as the methods used are approved.
Without this option, an admin would have to manually approve the whole script each time it changes.
https://support.cloudbees.com/hc/en-us/articles/207406207-Avoid-script-approvals-with-a-Jenkins-Pipeline-Groovy-script

farosas (approved):

PR #78, opened 2017-05-29
Playbook improvements prior to baremetal provisioning impl.

farosas:
Some early changes that alter current playbooks. This is mainly aimed at increasing modularity so that the following playbooks don't repeat code.


ansible/roles/jenkins-master/meta/main.yaml
```diff
@@ -1,3 +1,3 @@
 ---
 dependencies:
-  - role: handlers-jenkins-master
+  - role: nginx
```
olavph, line 3: I believe this is not needed. `restart nginx` is not called in `jenkins-master` role

olavph: `jenkins-seed-job` and the role recently created `jenkins-pipeline` do not have this file with `handlers-jenkins-master` as a dependency. I'll fix it in my PR, please remind me to update it if yours is merged first.

PR #79, opened 2017-06-06
Baremetal deployment

farosas:


farosas:
Work is almost done, I just need to fix some issues with the DHCP server and tune the timing parameters, but that shouldn't alter the patch-set too much.

ansible/roles/pkg_install/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Install distro packages
+  package: name={{item}} state=present
+  with_items:
+    "{{ distro_packages }}"
```
olavph, line 5: Where are those variables defined?

ansible/roles/dnsmasq/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install dnsmasq
+  package:
+    name=dnsmasq state=present
+  tags:
+    - setup
+
+- name: Create dnsmasq.d
+  file:
+    path=/etc/dnsmasq.d state=directory owner=root group=root mode=0755
```
olavph, line 10: Why is this empty directory required?

ansible/roles/deploy-baremetal/templates/kickstart
```diff
@@ -0,0 +1,44 @@
+# Created by Ansible on {{ansible_date_time.date}} at {{ansible_date_time.time}}.
```
olavph, line 1: Our playbooks (and Ansible examples) use `.j2` as templates extension. Emacs recognizes that extension as well.

ansible/roles/pkg_install/tasks/main.yaml
farosas: In the role that added this role as a dependency.

ansible/roles/dnsmasq/tasks/main.yaml
farosas: This is a directory for drop-in configuration files from other services.

http://blog.siphos.be/2013/05/the-linux-d-approach/

ansible/roles/deploy-baremetal/templates/kickstart
farosas: Ok, but first tell me what is the criteria for using .j2:

```
$ ls roles/*/templates
roles/builds/templates:
github_ssh_config.j2

roles/jenkins-master/templates:
admin_config.xml.j2  jenkins.conf  jenkinsd

roles/jenkins-seed-job/templates:
seed_job_config.xml.j2

roles/nginx/templates:
nginx.conf

roles/systemd/templates:
jenkins.service
```
At first I took a quick look and thought we we're not using the extension.

olavph: I thought all of them had it. Whatever then, I don't have a strong reason for any side.
We should fix this inconsistency later.

ansible/roles/baremetal-ctrl/meta/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+dependencies:
```
olavph, line 2: Just a warning. I frequently run playbooks with `--start-at` to execute only some of the later roles and save some time. If I added dependencies to those playbooks, the chain of dependencies would also be executed, preventing me from doing only what I wanted.
In other words, with this, you won't be able to execute this role by itself, even if the dependencies have been previously run.

ansible/roles/pkg_install/tasks/main.yaml
```diff
@@ -3,12 +3,20 @@
...
+    extra_args: --upgrade
+  when: python_packages
   tags: setup
 
 - name: Install Python packages
   pip: name={{item}} state=present
   with_items:
     "{{ python_packages }}"
-  when: "{{ python_packages }}"
+  when: python_packages
```
olavph: Those changes should be in the previous commit.

ansible/roles/deploy-baremetal/tasks/prepare.yaml
```diff
@@ -0,0 +1,65 @@
...
+  template:
+    src: pxeconfig-network
+    dest: "{{ http_server.root }}/pxelinux.cfg/01-{{ baremetal.mac_address | regex_replace(':', '-') }}"
+    owner: root
+    group: root
+    mode: 0644
+
+- name: Set machine_dir variable
+  set_fact:
+    machine_dir: "{{ http_server.root }}/{{ baremetal.mac_address }}"
```
olavph: Can't this variable be set in `deplay-baremetal/defaults/main.yaml`?
Also, you're using the `http_server` variable from `baremetal-ctrl/defaults/main.yaml`, shouldn't it be in a "global" vars file, or maybe that other role be a dependency of this one?

olavph (changes requested):

ansible/roles/deploy-baremetal/tasks/prepare.yaml
farosas: I dislike having things that will (almost) never change in vars.yaml because 1) people will want to change it for no reason 2) It creates confusion about which are the variables required for a specific setup and which are simply defaults.

The role cannot be a dependency of this one, because it is meant to be run only once with the controller node as a target, whereas the 'deploy' tag could be run multiple times.

Maybe we could use Ansible's group_vars to keep this more tidy. I'll explore with this a bit.

ansible/roles/dnsmasq/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+  tags:
+    - setup
+
+- name: Place dnsmasq.conf
+  template:
+    src=dnsmasq.conf dest=/etc/dnsmasq.conf
+    owner=root group=root mode=0644
+  tags:
+    - setup
```
mopsfelder: I believe dnsmasq service needs to be restarted (`notify: restart dnsmasq`) if dnsmasq.conf changes, doesn't it?

ansible/roles/dnsmasq/templates/dnsmasq.conf
```diff
@@ -0,0 +1,28 @@
+# Configuration file for dnsmasq.
```
mopsfelder, line 6: What about also adding the following header?

```
# /etc/dnsmasq.conf
#
# Installed by Ansible on {{ansible_date_time.date}} at {{ansible_date_time.time}}.
# WARNING: Any change to this file can be lost!
```

ansible/roles/baremetal-ctrl/meta/main.yaml
farosas: Well, these are just package dependencies, they shouldn't be interfering with the execution to the point that we would want to avoid them while debugging. But I get what you mean, I'll keep it in mind.

olavph (approved):

Issue #44, opened 2017-02-28
Automate LDAP configuration for Jenkins authentication


mopsfelder:
It may be convenient to users to have Jenkins authenticating against company's directory.

We can make LDAP configuration optional in the playbooks by using a control variable in `vars.yaml`, e.g. `ldap_config: yes`.

The following parameters can serve as a baseline and they can be added to `vars.yaml`:

```yaml
ldap_config: yes
ldap_server: "directory.company.com"
ldap_allow_blank_rootdn: yes
ldap_search_base: "ou=directory,o=company.com"
ldap_search_filter: "mail={0}"
```

We can have a task in `roles/jenkins-master/tasks/main.yaml` that only includes `roles/jenkins-master/tasks/ldap_config.yaml` when the variable `ldap_config` is `True`/`Yes`.

olavph:
Closed by #80

PR #80, opened 2017-06-08
Add role to configure Jenkins LDAP authentication

olavph:
This role is optional, because not all users may want to have LDAP
authentication. A simpler authentication model may also be useful in
test environments.


farosas (approved):

olavph:
Closes #44

PR #84, opened 2017-06-23
Disable caches and bind mounts when building old versions

olavph:
This prevents conflicts with concurrent (recent) builds at the expense
of longer (old) build times.


farosas (approved):

olavph:
Relates to https://github.com/open-power-host-os/builds/pull/272

PR #86, opened 2017-06-27
Allow customization of the periodic pipelines branch

farosas:


jenkins_jobs/periodic_pipelines.groovy
```diff
@@ -6,7 +6,7 @@ def createPeriodicPipeline(String name) {
 node('master') {
   dir('infrastructure') {
     git(url: 'https://github.com/$GITHUB_ORGANIZATION_NAME/infrastructure.git',
-        branch: 'master')
+        branch: '$REPOSITORY_COMMIT')
```
olavph, line 9: How would you set this variable?

farosas: The same way `GITHUB_ORGANIZATION_NAME`, via the seed job parameters. Is that wrong?

EDIT: It is already being used for the seed-job to download the jobs definitions

olavph: Ok, I didn't realize it was an existing parameter.
Just please make sure you mention this is used for periodic builds in the [parameter description](https://github.com/open-power-host-os/infrastructure/blob/master/ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2#L20).
There were cases that I've executed the seed job with a custom branch prior to merging it, and this would possibly break periodic builds when the master is later updated. This is something I shouldn't do anyway.

olavph (approved):

PR #87, opened 2017-06-30
Add slack notifications

farosas:


pipeline/lib/utils.groovy
```diff
@@ -0,0 +1,38 @@
...
+}
+
+def removeCredentials(String credentialsId) {
+  domain = Domain.global()
+
+  credentials = CredentialsMatchers.filter(
+    getCredentialsStore().getCredentials(domain),
+    CredentialsMatchers.withId(credentialsId))
+
+  getCredentialsStore().removeCredentials(domain, credentials.get(0))
```
olavph: What is returned by `credentials.get(0)`? Maybe use a intermediary variable to describe it.

pipeline/daily/stages.groovy
```diff
@@ -108,4 +109,35 @@ def createSymlinks() {
   utils.rsyncUpload('--links latest', PERIODIC_BUILDS_DIR_RSYNC_URL)
 }
 
+def shouldNotifyOnFailure() {
+  shouldNotify = true
+
+  for (param in params) {
+    if (param.key.startsWith("SLACK") && !param.value) {
+      shouldNotify = false
+      break
```
olavph, line 118: You could just `return false` here and `return true` outside the for, removing this variable. Up to you.

pipeline/daily/stages.groovy
```diff
@@ -108,4 +109,35 @@ def createSymlinks() {
...
+  String recipient = params.SLACK_NOTIFICATION_RECIPIENT
+  String failureMsg = "Daily build failed. ${BUILD_URL}console"
+
+  String tokenId = utils.addSecretString(
+    params.SLACK_TOKEN, "Slack token for $recipient @ $teamDomain")
+
+  try {
+    withCredentials([string(credentialsId: tokenId, variable: 'token')]){
+      slackSend(channel: recipient, teamDomain: teamDomain, token: token,
+		color: 'danger', message: failureMsg)
```
olavph, line 136: 'Danger' is my favorite color.

pipeline/daily.groovy
```diff
@@ -4,34 +4,43 @@ pipelineStages = load 'infrastructure/pipeline/daily/stages.groovy'
...
+	  }
+
+	  stage('Create symlinks') {
+	    pipelineStages.createSymlinks()
+	  }
+	}
+      }
+    } catch (Exception exception) {
+      if (pipelineStages.shouldNotifyOnFailure()) {
+	stage('Notify failure') {
```
olavph: I don't know if creating a stage for this is a good idea. If I recall correctly some tests I've made, this will make Jenkins workflow view lose information about stages that were not run.
For example, if it fails on 'Build packages', it will now show a column for 'Notify failure' and drop the columns for 'Build ISO', 'Upload build artifacts' and 'Create symlinks' from previous builds.

pipeline/daily/stages.groovy
farosas: It's intentional to avoid multiple return statements.

pipeline/daily.groovy
farosas: Ok, I'll play a bit with this and remove the stage if needed.

olavph (approved):

PR #94, opened 2017-07-18
Create a directory for keeping build information in the upload server

farosas:
This allows more information to be added later without having to download and reupload the STATUS file. Later stages of the pipeline may simply drop a .json file in the directory.


farosas:
This depends on Jenkins changes (/etc/jenkins/pipeline_constants.groovy) so it will not finish building.

olavph (approved):

PR #101, opened 2017-08-15
Do not build when there are no updates in release pipeline

olavph:
The pipeline will stop prematurely with a success if no updates are
detected.

<cde:info> Mirrored with LTC bug #157737 </cde:info>

### DIFF
--- a/pipeline/daily/stages.groovy
+++ b/pipeline/daily/stages.groovy
@@ -16,6 +16,8 @@ buildStages = load 'infrastructure/pipeline/build/stages.groovy'
 @Field String BUILDS_REPO_NAME
 @Field String COMMIT_BRANCH
 
+@Field Boolean hasUpdates = false
+
 def initialize(Map pipelineParameters = pipelineParameters,
                String triggerExpression = (
                  constants.NIGHTLY_BUILDS_CRON_EXPRESSION),
@@ -53,7 +55,7 @@ def updateVersions() {
   dir('builds') {
     git(url: "ssh://git@github.com/$params.GITHUB_ORGANIZATION_NAME/builds.git",
         branch: params.BUILDS_REPO_REFERENCE)
-    sh """\
+    exitCode = sh(script: """\
 python host_os.py    \
        --verbose \
        --work-dir $params.BUILDS_WORKSPACE_DIR \
@@ -64,7 +66,15 @@ python host_os.py    \
            --updater-email $params.GITHUB_BOT_EMAIL \
            --push-repo-url $VERSIONS_PUSH_REPO_URL \
            --push-repo-branch $COMMIT_BRANCH \
-"""
+""", returnStatus: true)
+
+    Integer SUCCESS_EXIT_CODE = 0
+    Integer NO_UPDATES_EXIT_CODE = 56
+    if (exitCode == SUCCESS_EXIT_CODE) {
+      hasUpdates = true
+    } else if (exitCode != NO_UPDATES_EXIT_CODE) {
+      error('Packages update failed')
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


pipeline/daily/stages.groovy
```diff
@@ -64,7 +66,17 @@ python host_os.py    \
            --updater-email $params.GITHUB_BOT_EMAIL \
            --push-repo-url $VERSIONS_PUSH_REPO_URL \
            --push-repo-branch $COMMIT_BRANCH \
-"""
+""", returnStatus: true)
+
+    Integer SUCCESS_EXIT_CODE = 0
+    Integer NO_UPDATES_EXIT_CODE = 56
```
mopsfelder, line 72: Do you think it makes sense to place these constants on a lib?

pipeline/daily/stages.groovy
```diff
@@ -64,7 +66,17 @@ python host_os.py    \
...
+    if (exitCode == SUCCESS_EXIT_CODE) {
+      updateSuccessful = true
+    } else if (exitCode == NO_UPDATES_EXIT_CODE) {
+      updateSuccessful = false
+    } else {
+      error('Packages update failed')
+    }
```
mopsfelder: The initial value of updateSuccessful is false, I think there is no need to set it again. What about the following?

```groovy
if (exitCode == SUCCESS_EXIT_CODE) {
  updateSuccessful = true
}

if (!updateSuccessful) {
  error('Packages update failed')
}
```

pipeline/devel.groovy
```diff
@@ -7,7 +7,7 @@ pipelineStages = load 'infrastructure/pipeline/daily/stages.groovy'
 pipelineParameters = load 'infrastructure/pipeline/devel/parameters.groovy'
 
 
-def execute() {
+def execute(Boolean skipWithoutUpdates = false) {
```
mopsfelder: The name skipWithoutUpdates gives me the idea that the pipeline will skip without touching or updating anything. I would have called it skipIfNoUpdates.

pipeline/daily/stages.groovy
olavph: We don't want to raise an error when NO_UPDATES_EXIT_CODE is returned. We don't want to do anything in that case. I could just leave an `else if (exitCode != NO_UPDATES_EXIT_CODE)` to raise the error.

pipeline/devel.groovy
olavph: Ok

pipeline/daily/stages.groovy
olavph: This is the only place those are used so far. If we start using them elsewhere, it might be a good idea to create a `returncodes.groovy` file.

pipeline/daily/stages.groovy
```diff
@@ -64,7 +66,15 @@ python host_os.py    \
            --updater-email $params.GITHUB_BOT_EMAIL \
            --push-repo-url $VERSIONS_PUSH_REPO_URL \
            --push-repo-branch $COMMIT_BRANCH \
-"""
+""", returnStatus: true)
+
+    Integer SUCCESS_EXIT_CODE = 0
+    Integer NO_UPDATES_EXIT_CODE = 56
+    if (exitCode == SUCCESS_EXIT_CODE) {
+      updateSuccessful = true
```
farosas: I feel this makes things confusing down the line. Can't we invert the logic and set something like `hasUpdates` instead?

farosas (changes requested):

pipeline/daily/stages.groovy
olavph: Yeah, that's better. There's no logic to invert, renaming the variable is enough.

farosas (approved):

PR #103, opened 2017-08-16
ansible: Add new `avocado` role

mopsfelder:
This role sets up avocado on a bare-metal system running CentOS 7.3.

### DIFF
--- a/ansible/roles/avocado/tasks/main.yaml
+++ b/ansible/roles/avocado/tasks/main.yaml
@@ -1,0 +1,59 @@
+- name: Setup avocado-vt configuration
+  copy:
+    src: vt.conf
+    dest: /etc/avocado/conf.d/vt.conf
+    owner: root
+    group: root
+    mode: 0644
+  notify: vt-bootstrap
+  tags:
+    - setup
+
+# This is necessary to create virbr0 interface. You can create it
+# manually, but libvirt does the job for you.
+- name: Start libvirtd service
+  service:
+    name: libvirtd
+    state: started
+    enabled: yes
+  tags:
+    - setup
+
+- name: Creating avocado directories
+  file:
+    path: "{{item}}"
+    state: directory
+    owner: root
+    group: root
+    mode: 0755
+  with_items:
+    - "{{avocado_conf_dir}}"
+    - "{{avocado_data_dir}}"
+    - "{{avocado_iso_dir}}"
+    - "{{avocado_log_dir}}"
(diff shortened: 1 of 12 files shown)


ansible/bvt-slave.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Setup slave node for Build Verification Tests
```
farosas: Let's remove all mentions to 'slave', this is a Jenkins-related concept. If I understand correctly, we will have to run both 'jenkins-slave' and 'bvt-slave' playbooks on a host to be able to use it in the pipeline, so I'd leave the "slave" setup to 'jenkins-slave' and use 'bvt-host' for this one.

ansible/roles/avocado/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+  package: name={{item}}
+  notify: vt-bootstrap
+  with_items:
+    - avocado
+    - avocado-plugins-vt
+    - libvirt
+    - qemu-kvm
+    - wget
+  tags:
+    - setup
```
farosas: I have used the `pkg_install` role for this in the baremetal setup, Could you take a look and consider using it as well? The main difference is that the list of deps is moved to defaults/main.yaml and the `pkg_install` goes into the `dependencies` field of this role. See:

The pkg_install role containing the boilerplate for installing packages:
https://github.com/open-power-host-os/infrastructure/blob/master/ansible/roles/pkg_install/tasks/main.yaml

And its usage:

The listing of the package dependencies that a role needs installed:
https://github.com/open-power-host-os/infrastructure/blob/master/ansible/roles/baremetal-ctrl/defaults/main.yaml#L5

The role pulling in `pkg_install` to do the work:
https://github.com/open-power-host-os/infrastructure/blob/master/ansible/roles/baremetal-ctrl/meta/main.yaml#L4

farosas (changes requested):

ansible/roles/avocado/tasks/main.yaml
mopsfelder: Sure thing! I'm updating this pull.

mopsfelder: As a side effect of this, I had to add a new role avocado-repo to configure avocado.repo before pkg_install role is called in avocado role. Otherwise, when pkg_install runs, it does not find avocado packages because avocado.repo is not set yet.

olavph (approved):

PR #106, opened 2017-08-23
Copy host keys to known_hosts in master and slave

YasminBeatriz:
Automate host key storage for GitHub and upload server in jenkins nodes.

- Add new task for `ansible/roles/ssh/tasks/main.yaml` that copies a `known_hosts` file containing GitHub's and upload server's host keys from source host to target hosts.

- In `ansible/jenkins_slave.yaml`, prompt the user whether (s)he wants to keep the default file path or provide a new one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,16 +94,16 @@ After the SSH proxy is configured, simply use the following to connect to the sy
 
 To automate SSH authentication between 2 hosts, it is necessary to:
  - create an SSH key pair if you still do not have one
- - add the target host to ~/.ssh/known_hosts in the source host. It is also
-   possible to do it by executing a simple SSH command from the source host to
-   the target host. Alternatively, you could disable "StrictHostKeyChecking"
-   for all SSH connections in the source host, but that may be a security hazard.
  - add the SSH private key to ~/.ssh/ in the source host
  - add the SSH public key to ~/.ssh/authorized_keys file in the target host
 
 To be able to upload the results of builds to a server and push commits to
 GitHub automatically, you must automate SSH authentication from Jenkins slaves
-to those two servers. The private SSH key is already installed in Jenkins nodes
+to those two servers. To have this being done automatically at the setup, make sure you have 
+both upload server and GitHub host keys stored in a local `~/.ssh/known_hosts` file. For 
+production servers, it is recommended that you provide a separate file with only 
+those two keys.
+The private SSH key is already installed in Jenkins nodes
 by Jenkins Ansible playbooks, so you do not need to do it manually.
 
 ### Automatically setup Jenkins master and slave(s) using Ansible playbooks
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -43,5 +43,23 @@ to enter the SSH password:
 The `--tags` option tells Ansible to only execute tasks with the specified tag.
 In the previous examples, only tasks with the `setup` tag would be executed.
 
-Provide the data requested by the playbook (eg Jenkins admin user name/password
-and SSH keys locations) and wait automatic setup finish.
+### List of manual input data
+
+Some Ansible playbooks will prompt you to input data to setup Jenkins master and slave nodes.
+
+- `jenkins-master.yaml` playbook
+
+	- Username and password to access Jenkins web UI with administrative privileges. Note that
+a passwordless `jenkins` user will also be created in master and slave nodes.
+	- Paths to SSH private and public keys used to communicate between master and slave nodes.
+
+- `jenkins-slave.yaml` playbook
+
+	- Paths to SSH private and public keys used to communicate between master and slave nodes.
+	- Path to the SSH private key used to upload 
+artifacts to the configured remote server.
+	- Path to the SSH private key used to push commits to GitHub.
+	- Path to the SSH known keys for remote hosts, usually `~/.ssh/known_hosts`. 
+The local file is just copied to remote target.
+
+Provide the data requested by the playbooks and wait automatic setup to finish.
--- a/ansible/jenkins-slave.yaml
+++ b/ansible/jenkins-slave.yaml
@@ -33,3 +33,7 @@
       prompt: "Enter GitHub user's private SSH key file path"
       default: "~/.ssh/open-power-host-os-builds-bot_id_rsa"
       private: no
+    - name: "known_hosts_file_path"
+      prompt: "Enter path to file containing known keys for GitHub and upload server hosts"
+      default: "~/.ssh/known_hosts"
+      private: no
--- a/ansible/roles/builds/tasks/main.yaml
+++ b/ansible/roles/builds/tasks/main.yaml
@@ -29,3 +29,10 @@
     owner={{builder_user_name}} group={{builder_user_name}} mode=0600
   tags:
     - setup
+
+- name: Add known keys for remote hosts
+  copy:
+    src={{known_hosts_file_path}} dest="{{jenkins_home_dir}}/.ssh/known_hosts"
+    owner={{builder_user_name}} group={{builder_user_name}}
+  tags:
+    - setup


ansible/roles/ssh/tasks/main.yaml
```diff
@@ -22,5 +22,14 @@
     - { src: "{{jenkins_public_ssh_key_file_path}}",
         dest: "{{jenkins_home_dir}}/.ssh/authorized_keys",
         owner: jenkins, group: jenkins, mode: "0600" }
+
+- name: Add Github and upload server's host key
+  copy:
+    src={{item.src}} dest={{item.dest}}
+    owner={{item.owner}} group={{item.group}}
+  with_items:
```
olavph: I'd place the values directly in `copy` since there's only one item, but that's a matter of taste. Feel free to keep as is.

ansible/jenkins-slave.yaml
```diff
@@ -33,3 +33,7 @@
       prompt: "Enter GitHub user's private SSH key file path"
       default: "~/.ssh/open-power-host-os-builds-bot_id_rsa"
       private: no
+    - name: "known_hosts_file_path"
+      prompt: "Enter upload server's host key file path"
```
olavph: I'd be more descriptive in this prompt string: "Enter path to file containing known keys for GitHub and upload server hosts"

ansible/roles/ssh/tasks/main.yaml
```diff
@@ -22,5 +22,14 @@
     - { src: "{{jenkins_public_ssh_key_file_path}}",
         dest: "{{jenkins_home_dir}}/.ssh/authorized_keys",
         owner: jenkins, group: jenkins, mode: "0600" }
+
+- name: Add Github and upload server's host key
```
olavph: This task has no control over whether the file will in fact contain the GitHub and upload server keys. This is up to the user and should be mentioned in the prompt and in the documentation.
I'd name it: "Add known keys for remote hosts" (using a similar wording as in https://en.wikibooks.org/wiki/OpenSSH/Client_Configuration_Files#.7E.2F.ssh.2Fknown_hosts)

olavph (changes requested):

ansible/roles/ssh/tasks/main.yaml
YasminBeatriz: So you suggest that I change to "Add known keys for remote hosts" in the task name, then be more specific in the Ansible `README` file about the content of `known_hosts`?

olavph: I think we can remove this paragraph from the README: "add the target host to ~/.ssh/known_hosts in the source host. It is also possible to do it by executing a simple SSH command from the source host to the target host. Alternatively, you could disable "StrictHostKeyChecking" for all SSH connections in the source host, but that may be a security hazard."

The prompt should be descriptive enough so the user knows what to do.

README.md
```diff
@@ -94,16 +94,16 @@ After the SSH proxy is configured, simply use the following to connect to the sy
...
  - add the SSH public key to ~/.ssh/authorized_keys file in the target host
 
 To be able to upload the results of builds to a server and push commits to
 GitHub automatically, you must automate SSH authentication from Jenkins slaves
-to those two servers. The private SSH key is already installed in Jenkins nodes
+to those two servers. This will be done automatically at the setup if you have both 
+upload server and GitHub's host keys stored in a local `~/.ssh/known_hosts` file. For 
+production servers, it is recommended that you provide a separate file with only 
+those two keys.
+The private SSH key is already installed in Jenkins nodes
```
YasminBeatriz: I think this sentence starting with "The private SSH key is already installed in Jenkins nodes..." should also be removed, or placed somewhere else with more context.

ansible/README.md
```diff
@@ -43,5 +43,15 @@ to enter the SSH password:
 The `--tags` option tells Ansible to only execute tasks with the specified tag.
 In the previous examples, only tasks with the `setup` tag would be executed.
 
-Provide the data requested by the playbook (eg Jenkins admin user name/password
-and SSH keys locations) and wait automatic setup finish.
+For Jenkins master playbook you only need to provide a jenkins username and password in 
+the target host.
```
olavph: Not true, there are four prompts in the master playbook: https://github.com/open-power-host-os/infrastructure/blob/master/ansible/jenkins-master.yaml#L21

ansible/README.md
```diff
@@ -43,5 +43,15 @@ to enter the SSH password:
...
+
+Jenkins slave playbook will prompt you to provide the following data:
+- Jenkins private SSH key file path _(press Enter to keep default value)_
+- Jenkins public SSH key file path _(press Enter to keep default value)_
+- Upload server user's private SSH key file path _(enter your SSH private key)_
+- GitHub user's private SSH key file path _(enter your SSH private key)_
+- Path to file containing known keys for GitHub and upload server hosts _(press Enter to keep 
+default known_hosts file path or enter any alternate file you may have)_
```
olavph: I don't think the comments in parentheses help much. I'd explain better what the keys are used for:
Master:
- Jenkins admin user name and password: used to access the Jenkins UI with administrative privileges.
- Jenkins private and public SSH key file path: used to communicate between the master and the slave nodes.

Slave:
- Jenkins private and public SSH key file path: used to communicate between the master and the slave nodes.
- Upload server user's private SSH key file path: used to upload builds to the configured server.
- GitHub user's private SSH key file path: used to push commits to GitHub in periodic pipelines.
- Path to file containing known keys for GitHub and upload server hosts: used to recognize the server identities.

ansible/roles/ssh/tasks/main.yaml
```diff
@@ -22,5 +22,11 @@
     - { src: "{{jenkins_public_ssh_key_file_path}}",
         dest: "{{jenkins_home_dir}}/.ssh/authorized_keys",
         owner: jenkins, group: jenkins, mode: "0600" }
+
+- name: Add known keys for remote hosts
+  copy:
+    src={{known_hosts_file_path}} dest="{{jenkins_home_dir}}/.ssh/known_hosts"
+    owner=jenkins group=jenkins
+
```
olavph: This role is executed by both master and slave playbooks and will fail on the master because there is no value for `known_hosts_file_path` when it is executed.
Move it to the end of the [builds role](https://github.com/open-power-host-os/infrastructure/blob/master/ansible/roles/builds/tasks/main.yaml), which is where the SSH keys for those servers are actually configured.

olavph (changes requested):

ansible/README.md
YasminBeatriz: Oops, my mistake.

ansible/README.md
YasminBeatriz: It is certainly good to explain what each key means, but that still doesn't explain which ones are automatically generated and which ones are user generated (therefore with a pre-existing path provided by the user).

ansible/README.md
```diff
@@ -43,5 +43,20 @@ to enter the SSH password:
...
+
+Slave:
+- Jenkins private and public SSH key file path: used to communicate between the master 
+and the slave nodes.
+- Upload server user's private SSH key file path: used to upload builds to the configured server.
+- GitHub user's private SSH key file path: used to push commits to GitHub in periodic pipelines.
+- Path to file containing known keys for GitHub and upload server hosts: used to recognize 
+the server identities. 
+
+Provide the data requested by the playbooks and wait automatic setup finish.
```
YasminBeatriz: I've applied the changes requested by @olavph (see [outdated comment](https://github.com/open-power-host-os/infrastructure/pull/106#discussion_r135278310)) 
I'd like to point out that it is good, indeed, to describe what each key means, but during the execution of the playbooks I was still confused as to which ones are automatically generated and which ones are user generated.

For example, in the slave playbook, the first item (Jenkins private and public SSH key) is automatically generated by the playbook during setup, so the user can just press Enter and proceed, while the following 2 items (Upload server and GitHub private SSH keys) refers to a key manually generated by the user in the previous steps (assuming the user is following the setup from README in infrastructure main page) so in the prompt the they will have to overwrite the default value. However, this info on what action the user should take in each case is not clear in this tutorial.

mopsfelder: I'd be more specific here, listing playbook and variable names. Something like:

Some Ansible playbooks will prompt you to input data to setup Jenkins master and slave nodes.

- `jenkins-master.yaml` playbook
  - `jenkins_admin_user` and `jenkins_admin_password`: Username and password to access Jenkins web UI with administrative privileges.
  - `jenkins_private_ssh_key_file_path` and `jenkins_public_ssh_key_file_path`: Paths to SSH private and public keys used to communicate between master and slave nodes.

- `jenkins-slave.yaml` playbook
  - `jenkins_private_ssh_key_file_path` and `jenkins_public_ssh_key_file_path`: Paths to SSH private and public keys used communicate between master and slave nodes.
  - `upload_server_user_private_ssh_key_file_path`: Path to the SSH private key used to upload artifacts to the configured remote server.
  - `github_user_private_ssh_key_file_path`: Path to the SSH private key used to push commits to GitHub.
  - `known_hosts_file_path`: Path to the SSH known keys for remote hosts, usually `~/.ssh/known_hosts`. The local file is just copied to remote target.

olavph: @mopsfelder I do not recommend using the variable names, since the prompts that appear on execution show a phrase instead of them (i.e. `Enter Jenkins private SSH key file path [~/.ssh/jenkins_id_rsa]:`). This makes it a bit harder to associate the prompts with this documentation.

olavph: @YasminBeatriz None of the SSH keys are automatically generated. They must be previously created by the user and may be anywhere in its system (although a standard path is suggested by default).

ansible/roles/builds/tasks/main.yaml
```diff
@@ -29,3 +29,10 @@
     owner={{builder_user_name}} group={{builder_user_name}} mode=0600
   tags:
     - setup
+
+- name: Add known keys for remote hosts
+  copy:
+    src={{known_hosts_file_path}} dest="{{jenkins_home_dir}}/.ssh/known_hosts"
+    owner=jenkins group=jenkins
```
mopsfelder: I think this should be

```
owner={{builder_user_name}} group={{builder_user_name}}```

mopsfelder (approved):

olavph (approved):
You'll need a rebase on top of master.

mopsfelder (approved):

PR #113, opened 2017-09-20
Parameterize Github URL at infrastructure setup

YasminBeatriz:
Instead of having GitHub host name (github.com) hard-coded at the setup, now replace it with `$GIT_REPOS_BASE_URL` so it allows users to use enterprise instances of GitHub as well (github.ibm.com, for example).

### DIFF
--- a/ansible/roles/builds/templates/github_ssh_config.j2
+++ b/ansible/roles/builds/templates/github_ssh_config.j2
@@ -1,4 +1,4 @@
-host github.com
- HostName github.com
+host github
+ HostName {{pipeline_constants.GITHUB_DOMAIN}}
  IdentityFile "{{builder_home_dir}}/.ssh/github_id_rsa"
  User git
--- a/ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
+++ b/ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
@@ -13,7 +13,7 @@
         <hudson.model.StringParameterDefinition>
           <name>REPOSITORY_URL</name>
           <description>Repository to download the DSL job descriptors.</description>
-          <defaultValue>ssh://git@github.com/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
+          <defaultValue>ssh://git@github/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.StringParameterDefinition>
           <name>REPOSITORY_COMMIT</name>
--- a/ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
+++ b/ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
@@ -13,7 +13,7 @@
         <hudson.model.StringParameterDefinition>
           <name>REPOSITORY_URL</name>
           <description>Repository to download the DSL job descriptors.</description>
-          <defaultValue>https://github.com/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
+          <defaultValue>ssh://git@github.com/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.StringParameterDefinition>
           <name>REPOSITORY_COMMIT</name>
--- a/ansible/vars.yaml
+++ b/ansible/vars.yaml
@@ -1,5 +1,6 @@
 ---
 pipeline_constants:
+  GITHUB_DOMAIN: github.com
   GITHUB_ORGANIZATION_NAME:
   GITHUB_BOT_NAME:
   GITHUB_BOT_USER_NAME:
--- a/jenkins_jobs/build_old_host_os_versions.groovy
+++ b/jenkins_jobs/build_old_host_os_versions.groovy
@@ -5,7 +5,7 @@ job('build_old_host_os_versions') {
 		"${GITHUB_ORGANIZATION_NAME}",
 		'GitHub organization from where the Host OS repositories will be checked out.')
     stringParam('BUILDS_REPO_URL',
-		"https://github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
+		"ssh://git@github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
 		'URL of the builds repository.')
     stringParam('BUILDS_REPO_REFERENCE', 'origin/master',
 		'Git reference to checkout from the builds repository.')
--- a/jenkins_jobs/build_old_host_os_versions.groovy
+++ b/jenkins_jobs/build_old_host_os_versions.groovy
@@ -5,7 +5,7 @@ job('build_old_host_os_versions') {
 		"${GITHUB_ORGANIZATION_NAME}",
 		'GitHub organization from where the Host OS repositories will be checked out.')
     stringParam('BUILDS_REPO_URL',
-		"ssh://git@github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
+		"ssh://git@github/${GITHUB_ORGANIZATION_NAME}/builds.git",
 		'URL of the builds repository.')
     stringParam('BUILDS_REPO_REFERENCE', 'origin/master',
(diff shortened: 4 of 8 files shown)


ansible/roles/builds/templates/github_ssh_config.j2
```diff
@@ -1,4 +1,4 @@
-host github.com
- HostName github.com
+host github
+ HostName params.GITHUB_REPOS_BASE_URL
```
farosas: This doesn't look like it would work, am I missing something? `params` is a Jenkins variable isn't it?

pipeline/build/stages.groovy
```diff
@@ -47,7 +47,7 @@ def authorize() {
   } else {
     echo("Retrieving authorized users from GitHub organization " +
          params.GITHUB_ORGANIZATION_NAME)
-    String orgMembersURL = ("https://api.github.com/orgs/" +
+    String orgMembersURL = ("https://api.$params.GITHUB_REPOS_BASE_URL/orgs/" +
```
farosas: Please be consistent in the variable access syntax.

ansible/vars-bvt.yaml
```diff
@@ -12,6 +12,6 @@ avocado_conf: "{{avocado_conf_dir}}/avocado.conf"
 test_providers_dir: "{{avocado_data_dir}}/avocado-vt/test-providers.d/downloads"
 
 avocado_repos:
-  - { repo: "https://github.com/autotest/tp-libvirt.git", dest: "{{test_providers_dir}}/io-github-autotest-libvirt" }
-  - { repo: "https://github.com/autotest/tp-qemu.git", dest: "{{test_providers_dir}}/io-github-autotest-qemu" }
-  - { repo: "https://github.com/spiceqa/tp-spice.git", dest: "{{test_providers_dir}}/io-github-spiceqa-spice" }
+  - { repo: "https://{{pipeline_constants.GITHUB_REPOS_BASE_URL}}/autotest/tp-libvirt.git", dest: "{{test_providers_dir}}/io-github-autotest-libvirt" }
```
farosas: This is a bit confusing. "base url" looks like it would contain the _https://_ part of the URL too. What about `GITHUB_DOMAIN`?

farosas (changes requested):

pipeline/build/parameters.groovy
```diff
@@ -1,4 +1,7 @@
 Map pipelineParameters = [
+  GITHUB_DOMAIN: [
+    defaultValue: constants.GITHUB_DOMAIN,
+    description: 'GitHub instance.'],
```
farosas: "Domain of the GitHub instance (github.com, github.org.com)"

farosas (changes requested):

ansible/vars-bvt.yaml
```diff
@@ -12,6 +12,6 @@ avocado_conf: "{{avocado_conf_dir}}/avocado.conf"
 test_providers_dir: "{{avocado_data_dir}}/avocado-vt/test-providers.d/downloads"
 
 avocado_repos:
-  - { repo: "https://github.com/autotest/tp-libvirt.git", dest: "{{test_providers_dir}}/io-github-autotest-libvirt" }
-  - { repo: "https://github.com/autotest/tp-qemu.git", dest: "{{test_providers_dir}}/io-github-autotest-qemu" }
-  - { repo: "https://github.com/spiceqa/tp-spice.git", dest: "{{test_providers_dir}}/io-github-spiceqa-spice" }
+  - { repo: "https://{{pipeline_constants.GITHUB_DOMAIN}}/autotest/tp-libvirt.git", dest: "{{test_providers_dir}}/io-github-autotest-libvirt" }
+  - { repo: "https://{{pipeline_constants.GITHUB_DOMAIN}}/autotest/tp-qemu.git", dest: "{{test_providers_dir}}/io-github-autotest-qemu" }
+  - { repo: "https://{{pipeline_constants.GITHUB_DOMAIN}}/spiceqa/tp-spice.git", dest: "{{test_providers_dir}}/io-github-spiceqa-spice" }
```
olavph: Those are upstream projects and repositories that we do not modify. They should keep the fixed "github.com" domain.

olavph (changes requested):

farosas (approved):

olavph (approved):

ansible/vars.yaml
```diff
@@ -1,5 +1,6 @@
 ---
 pipeline_constants:
+  GITHUB_DOMAIN:
```
olavph: This should have a default value "github.com"

ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
```diff
@@ -13,7 +13,7 @@
         <hudson.model.StringParameterDefinition>
           <name>REPOSITORY_URL</name>
           <description>Repository to download the DSL job descriptors.</description>
-          <defaultValue>https://github.com/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
+          <defaultValue>https://{{pipeline_constants.GITHUB_DOMAIN}}/{{pipeline_constants.GITHUB_ORGANIZATION_NAME}}/infrastructure.git</defaultValue>
         </hudson.model.StringParameterDefinition>
```
olavph: GITHUB_DOMAIN needs to be added as a parameter in this file. You can place it right before GITHUB_ORGANIZATION_NAME

olavph (changes requested):

ansible/roles/jenkins-seed-job/templates/seed_job_config.xml.j2
olavph: The SSH key is not being used because the host `github.com` no longer matches the entry `github` from the config file. You'll have to replace every occurrence of `ssh://git@github.com/` to `ssh://git@github/`.
I'd also change every HTTPS URL to SSH, to make use of the same configuration and detect authorization issues earlier. With this, you can also remove my suggestion above and the variable `{{pipeline_constants.GITHUB_DOMAIN}}` from this file.

farosas (changes requested):
Let's avoid overwriting the user's `.ssh/config` file because the user might be a real one.

jenkins_jobs/build_old_host_os_versions.groovy
```diff
@@ -5,7 +5,7 @@ job('build_old_host_os_versions') {
 		"${GITHUB_ORGANIZATION_NAME}",
 		'GitHub organization from where the Host OS repositories will be checked out.')
     stringParam('BUILDS_REPO_URL',
-		"https://${GITHUB_DOMAIN}/${GITHUB_ORGANIZATION_NAME}/builds.git",
+		"ssh://git@github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
```
olavph: Remove `.com`

jenkins_jobs/build_old_host_os_versions.groovy
```diff
@@ -5,7 +5,7 @@ job('build_old_host_os_versions') {
 		"${GITHUB_ORGANIZATION_NAME}",
 		'GitHub organization from where the Host OS repositories will be checked out.')
     stringParam('BUILDS_REPO_URL',
-		"https://github.com/${GITHUB_ORGANIZATION_NAME}/builds.git",
+		"https://${GITHUB_DOMAIN}/${GITHUB_ORGANIZATION_NAME}/builds.git",
```
olavph: Changing to `${GITHUB_DOMAIN}` (github.com) breaks the build because the SSH configuration for github (whose name change to `github` only) is not used. The next commit should be squashed into this one, so we don't have broken/unnecessary changes polluting the git history.
(This comment should be in an SSH URL, sorry)

olavph (changes requested):

farosas (approved):

jenkins_jobs/build_old_host_os_versions.groovy
olavph: The variable GITHUB_DOMAIN does not exist here. It would need to be created in the seed job to be accessible.
Since we are changing the HTTP URLs to SSH, this would be a useless change, that would need to be reverted in the next commit. It makes more sense to reorder the commits so that the change from HTTP to SSH comes first (still using the fixed github.com domain) and this commit that parameterizes the GitHub URL comes second.

By doing this, the **only** place where GITHUB_DOMAIN should appear is in the SSH config file.

olavph (changes requested):

olavph (approved):

PR #115, opened 2017-09-28
Stop overwriting ssh config file on target host

YasminBeatriz:
Users may have other entries in their `~/.ssh/config` file that are not related to Ansible. It is not interesting to have them being overwritten by the playbooks in the target host.

The `template` module now has been replaced by `blockinfile` module, which appends the content of `github_ssh_config.j2` to the end of `~/.ssh/config` (if not already there).  If the file doesn't exist in the target host, it will create a new one. 

The `blockinfile` appends the text block surrounded by customizable marker lines.

### DIFF
--- a/ansible/roles/builds/tasks/main.yaml
+++ b/ansible/roles/builds/tasks/main.yaml
@@ -23,10 +23,13 @@
     - setup
 
 - name: Configure SSH to push to GitHub repository
-  template:
-    src=github_ssh_config.j2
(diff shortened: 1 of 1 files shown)


mopsfelder (commented):
~~What if we use the known_hosts Ansible module to accomplish this?
http://docs.ansible.com/ansible/latest/known_hosts_module.html~~

Nevermind. I for some reason thought this was about known_hosts.

olavph (approved):

PR #117, opened 2017-10-18
Check for changes from previous build from release pipeline

olavph:


pipeline/build/stages.groovy
```diff
@@ -138,14 +138,9 @@ def buildPackages() {
...
-  dir('versions') {
-    // The commit hashes should be accessible from the GitSCM object
-    // https://issues.jenkins-ci.org/browse/JENKINS-34455
-    versions_repo_commit_id = sh(
-      script:"git rev-parse HEAD", returnStdout: true).trim()
-    echo versions_repo_commit_id
-  }
+  checkoutResult = utils.checkoutRepo('versions', gitRepos)
+  versionsRepoCommitId = checkoutResult['GIT_COMMIT']
+  echo(versionsRepoCommitId)
```
mopsfelder, line 143: Are we missing a `$` sign here?

olavph: You mean in the `echo` call? No, it only needs to be used to access variables inside strings, e.g.: `echo("$versionsRepoCommitId")`

mopsfelder (approved):

PR #121, opened 2017-10-27
ansible: Add role `time`

mopsfelder:
This role `time` is responsible for setting timezone and adjusting
time by running ntpdate command in Jenkins master and slave nodes.

### DIFF
--- a/ansible/roles/time/tasks/main.yaml
+++ b/ansible/roles/time/tasks/main.yaml
@@ -1,0 +1,34 @@
+---
+- name: Set timezone
+  timezone: name="{{ timezone }}"
+  notify: restart ntpd
+  tags:
+    - setup
+
+- name: Stop NTP service before running ntpdate
+  service: name=ntpd state=stopped
+  tags:
+    - setup
+
+- name: Configure ntp.conf
+  template:
+    src: ntp.conf.j2
+    dest: /etc/ntp.conf
+    owner: root
+    group: root
+    mode: 0644
+  notify: restart ntpd
+  tags:
+    - setup
+
+- name: Run ntpdate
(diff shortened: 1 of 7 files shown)


ansible/roles/time/templates/ntp.conf.j2
```diff
@@ -0,0 +1,57 @@
...
+#trustedkey 4 8 42
+
+# Specify the key identifier to use with the ntpdc utility.
+#requestkey 8
+
+# Specify the key identifier to use with the ntpq utility.
+#controlkey 8
+
+# Enable writing of statistics records.
+#statistics clockstats cryptostats loopstats peerstats
```
olavph: Don't you think it's better to remove these comments to keep the file cleaner?
Assuming the line `includefile /etc/ntp/crypto/pw` is unnecessary with `#crypto` disabled, lines 25-51 could be removed.

ansible/roles/time/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+- name: Stop NTP service before running ntpdate
+  service: name=ntpd state=stopped
+  tags:
+    - setup
+
+- name: Run ntpdate
+  command: "ntpdate -t 5 {{ item }}"
+  notify: restart ntpd
+  with_items:
+    - "{{ ntp_servers }}"
```
olavph: Just a curiosity, you are passing a nested list here. `with_items: "{{ntp_servers}}"` would be enough.
From https://github.com/ansible/ansible/issues/5913, ansible flattens the lists in `with_items`, so the result is the same.
As we already have both formats in our playbook, any one is ok.

olavph (approved):

PR #123, opened 2017-11-06
Rename build-iso to build-images

mopsfelder:
This renames build-iso to build-images in all pipelines; updates
function names; and change build-images to only build installable tree
with --install-tree option.

Depends on https://github.com/open-power-host-os/builds/pull/304

### DIFF
--- a/pipeline/build/stages.groovy
+++ b/pipeline/build/stages.groovy
@@ -198,16 +198,16 @@ python host_os.py \\
   }
 }
 
-def buildIso(boolean cleanWorkspace = true) {
+def buildImages(boolean cleanWorkspace = true) {
   if (!buildInfo.build_packages_finished) {
-    error('Skipping ISO build because packages build failed')
+    error('Skipping images build because packages build failed')
   }
 
-  buildInfo.build_iso_finished = false
+  buildInfo.build_images_finished = false
 
   if (triggeredRepoName) {
     utils.setGithubStatus(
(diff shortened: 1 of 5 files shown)


pipeline/build/stages.groovy
```diff
@@ -246,35 +246,36 @@ def buildIso(boolean cleanWorkspace = true) {
...
-        buildInfo.build_iso_finished = true
+        buildInfo.build_images_finished = true
       }
     }
   }
 
-  if (buildInfo.build_iso_finished) {
+  if (buildInfo.build_images_finished) {
     sh 'ln -s builds/result/iso/latest iso'
     stash name: 'iso_dir', includes: 'iso/'
```
olavph: I think the stash name should be called "images_dir" to be consistent with the other names. Do you think the default result dir should also be changed, or "iso" is a more relatable name to our users?
The dir name is set here, btw: https://github.com/open-power-host-os/builds/pull/304/files#diff-0a731305674a70cfdf5547367516c18fR85

mopsfelder: I'm renaming `iso_dir` to `images_dir` in stash calls.

As to `iso` being the resulting directory, it's fine, IMO. For example:

When only iso is built:

```iso/*.iso```

When only install tree is built:

```iso/os/```

When both iso and install tree are built:

```
iso/*.iso
iso/os/
```

olavph (approved):

PR #125, opened 2017-11-10
Create BVT script

olavph:


ansible/roles/avocado/files/host-os-bvt.ini
```diff
@@ -0,0 +1,9 @@
+[provider]
+uri: https://github.com/olavph/bvt.git
```
mopsfelder: What if we create a new repo under open-power-host-os org to host this code?

ansible/roles/avocado/files/host-os-bvt.ini
```diff
@@ -0,0 +1,9 @@
...
+branch: add-yum-update-test
+[generic]
+subdir: generic
+
+# A bug in avocado-vt prevents the tests from being listed without this
```
mopsfelder: Can we paste a link to the bug?

ansible/vars-bvt.yaml
```diff
@@ -9,11 +9,21 @@ avocado_test_dir: "{{avocado_dir}}/tests"
 avocado_conf_dir: "{{ansible_env.HOME}}/.config/avocado"
 avocado_conf: "{{avocado_conf_dir}}/avocado.conf"
 
-test_providers_dir: "{{avocado_data_dir}}/avocado-vt/test-providers.d/downloads"
+avocado_vt_repo: "https://github.com/olavph/avocado-vt.git"
```
mopsfelder: Same here. Perhaps creating a new repo under HostOS org may be useful.

ansible/vars-bvt.yaml
```diff
@@ -9,11 +9,21 @@ avocado_test_dir: "{{avocado_dir}}/tests"
...
-  - { repo: "https://github.com/autotest/tp-libvirt.git", dest: "{{test_providers_dir}}/io-github-autotest-libvirt" }
-  - { repo: "https://github.com/autotest/tp-qemu.git", dest: "{{test_providers_dir}}/io-github-autotest-qemu" }
-  - { repo: "https://github.com/spiceqa/tp-spice.git", dest: "{{test_providers_dir}}/io-github-spiceqa-spice" }
+  - { repo: "https://github.com/autotest/tp-libvirt.git",
+  dest: "{{test_providers_downloads_dir}}/io-github-autotest-libvirt" }
+  - { repo: "https://github.com/autotest/tp-qemu.git",
+  dest: "{{test_providers_downloads_dir}}/io-github-autotest-qemu" }
+  - { repo: "https://github.com/spiceqa/tp-spice.git",
+  dest: "{{test_providers_downloads_dir}}/io-github-spiceqa-spice" }
+  - { repo: "https://github.com/olavph/bvt.git",
```
mopsfelder: Same here.

ansible/roles/avocado/tasks/main.yaml
```diff
@@ -81,3 +81,13 @@
     - "{{avocado_repos}}"
   tags:
     - setup
+
+- name: Install BVT script
+  copy:
+    src: "{{playbook_dir}}/../scripts/host-os-bvt.py"
+    dest: /usr/bin/host-os-bvt
```
mopsfelder, line 122: What if this executable is installed under /usr/local/bin/host-os-bvt? I see no problem with /usr/bin, it would just be a clearer separation from what is from the distro and what is local settings.

scripts/host-os-bvt.py
```diff
@@ -0,0 +1,159 @@
+#!/bin/python2
```
mopsfelder: What if `#!/bin/env python2` was used instead? This would, for example, allow one to run this script under virtualenv.

scripts/host-os-bvt.py
```diff
@@ -0,0 +1,159 @@
...
+import logging
+import shlex
+import shutil
+import subprocess
+
+AVOCADO_RUN_COMMAND = (
+    "/usr/bin/avocado run --vt-type {vt_type} --vt-guest-os {vt_guest_os} "
+    "--failfast on {tests}")
+VT_TYPE = "libvirt"
+VT_GUEST_OS = "CentOS.7.4"
```
mopsfelder: Do you think this VT_GUEST_OS could be set via a command line option? This would, for example, allow one to specify a different guest OS and, in the future, when a new CentOS version is released (e.g.: 7.5), we could just update the parameter value, not changing the code.

scripts/host-os-bvt.py
```diff
@@ -0,0 +1,159 @@
...
+IMPORT_GUEST_TEST = ("io-github-autotest-qemu.unattended_install.import.import"
+                     ".default_install.aio_native")
+REMOVE_GUEST_TEST = "io-github-autotest-libvirt.remove_guest.without_disk"
+INSTALL_BASE_OS_TEST = ("io-github-autotest-qemu.unattended_install.url.http_ks"
+                        ".default_install.aio_native")
+UPDATE_BASE_OS_TEST = "io-github-autotest-qemu.yum_update"
+INSTALL_HOST_OS_TEST = "host-os-bvt.yum_install"
+UPDATE_HOST_OS_TEST = "host-os-bvt.yum_update"
+
+DISK_PATH = "/root/avocado/data/avocado-vt/images/centos74-ppc64le.qcow2"
```
mopsfelder: Do you think this disk path could also be parameterized?

mopsfelder (commented):
Thanks for putting all of this together!

Some of my comments are about making BVT script more flexible with more parameters, just minor enhancements that we can implement in the future, if we feel the need to do so.

Another point is about having those repositories under `olavph` migrated to `open-power-host-os` organization.

ansible/roles/avocado/files/host-os-bvt.ini
olavph: My intent was to fix the bug already, but I didn't have the time to do it. I will open it.

ansible/roles/avocado/tasks/main.yaml
olavph: This command is executed with sudo (because it is required by avocado with libvirt) and the root user does not have `/usr/local/bin` in the path by default, so I wanted to avoid the hassle of dealing with this. Do you think that's the proper way to do it?

ansible/vars-bvt.yaml
olavph: Will do.

ansible/vars-bvt.yaml
olavph: Will do.

scripts/host-os-bvt.py
olavph: Ok.

scripts/host-os-bvt.py
olavph: Yeah that's better, but requires a few changes in the code (or using global variables, which I dislike). Could we leave as an enhancement for later?

scripts/host-os-bvt.py
olavph: Yeah, it should. Could we leave as an enhancement for later?

ansible/roles/avocado/files/host-os-bvt.ini
olavph: Will do.

ansible/roles/avocado/files/host-os-bvt.ini
olavph: Done: https://github.com/avocado-framework/avocado-vt/issues/1222

ansible/roles/avocado/tasks/main.yaml
mopsfelder: My root's PATH is defined with `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin` and I don't recall changing anything specific to PATH. Let's not worry about that. It doesn't change any function behaviour.

scripts/host-os-bvt.py
mopsfelder: Sure, no problem.

scripts/host-os-bvt.py
mopsfelder: Sure.

PR #129, opened 2018-02-05
Stop failing the whole build when BVT fails

farosas:
The Build Verification Tests are a bit unstable right now due to external factors (network issues, OS issues, etc), so let's avoid failing the whole build when BVT fails. Passing the build allows us to focus on fixing the BVT issues instead of having to scramble to make sure the build gets re-executed properly. In the case that the BVT failure is an actual one - related to the build itself -, it's still better to push some artifacts (e.g. RPMs) to the repo anyway, since debugging and fixing the issue might not be done until the next build starts.


mopsfelder (approved):
Changes look sane to me. Thanks!

PR #133, opened 2018-08-03
Unset periodic pipelines

mopsfelder:


jenkins_jobs/periodic_pipelines.groovy
```diff
@@ -1,27 +0,0 @@
...
-      }
-    }
-    triggers {
-      cron(cronExpression)
-    }
-  }
-}
-
-createPeriodicPipeline('devel', "$NIGHTLY_BUILDS_CRON_EXPRESSION")
-createPeriodicPipeline('release', "$NIGHTLY_BUILDS_CRON_EXPRESSION")
```
farosas, line 27: I was thinking more of removing the `createPeriodicPipeline` line for devel and just disabling the release pipeline in the Jenkins interface. We will have to run the release pipeline from time to time won't we?

mopsfelder: That works too.

